Issue #25, opened 2018-02-20
Remove Windows XP and Vista support in themes and CSS


wolfbeast:
What is says on the tin.

- [ ] Remove Windows theme components and resources for pre-Win7
- [ ] Remove `@media()` queries from CSS for unsupported Windows versions.

PR #27, opened 2018-02-20
Remove Windows XP and Windows Vista theme components and resources.

wolfbeast:
This is part 1+2 for #25 
Tested briefly on Windows 7, but needs to be verified on Windows 8.0, 8.1 and 10.

@Lootyhoof maybe you can have a gander at this?


wolfbeast:
If merged, this resolves #25.

browser/themes/windows/browser.css
```diff
@@ -2641,19 +2521,14 @@ notification.pluginVulnerable > .notification-inner > .messageCloseButton {
   margin-top: -4px;
 }
 
-
-@media not all and (-moz-os-version: windows-xp) {
```
Lootyhoof: If this was the main blocker for the `browser-aero` code, can this now reside in `browser.css`?

browser/themes/windows/places/organizer.css
```diff
@@ -151,35 +139,28 @@
   padding-inline-end: 9px;
 }
 
-
-@media not all and (-moz-os-version: windows-xp) {
-  #placesView {
```
Lootyhoof: Where has this gone? We removed the explicit border drawing above, but does it add one regardless?

wolfbeast: Above under root view we set a border, but then removed it again on everything except XP. Since XP is gone, we no longer need the border being set either. Removing both blocks has the same effect as having 2 blocks adding+removing the same thing.

browser/themes/windows/browser.css
wolfbeast: I'm sure we can move browser-aero into the main browser css, but let's do that as a separate thing.

Lootyhoof:
I'd say this LGTM, then. Ship It!

PR #40, opened 2018-03-04
Port across devtools enhancements - 3

### LINKED ISSUE #31: Port across devtools enhancements
@janekptacijarabaci has done a lot of devtools work on Möbius, and this does not all graft cleanly (or at all) to UXP.
Since our base "catch-up" is pretty much done now, this can be put back in.

I'm assigning you to this, Janek -- could you please see about getting a working copy up with devtools changes ported across and **just make it a single (or a few) flattened PRs to get all this into our tree again**? No need to make a bunch of extra work here because it's related to a separate module, anyway, and would just mean a lot of overhead otherwise to make individual PRs for small changes (and would probably cause merge conflicts as well if so). Just make it the quickest way to get all this back into the devtools; listing the changes in the PR's description is enough.

janekptacijarabaci:
IMHO: This resolves #31.

---

Native in moebius:
[Bug 1302989: Make storage inspector work with file:// when # is in the URL](https://bugzilla.mozilla.org/show_bug.cgi?id=1302989)

---

[moebius#346: Storage Inspector should trim port from hosts for cookies](https://github.com/MoonchildProductions/moebius/pull/346)

---

Can I create another issue to create another PR for DevTools (without moebius)?
Currently I have three (small) fixes.

### DIFF
--- a/devtools/client/storage/test/browser.ini
+++ b/devtools/client/storage/test/browser.ini
@@ -9,6 +9,7 @@ support-files =
   storage-idb-delete-blocked.html
   storage-indexeddb-duplicate-names.html
   storage-listings.html
+  storage-listings-with-fragment.html
   storage-localstorage.html
   storage-overflow.html
   storage-search.html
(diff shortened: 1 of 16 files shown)


wolfbeast:
Feel free to create another issue + PR with devtools enhancements you have ready. I'll be closing #31 after this merge because the bulk import should be complete with this.

PR #42, opened 2018-03-04
Import lss code from esr52

trav90:
Fixes build bustage on Linux (even with crashreporter disabled).

Linux builds fail (regardless whether crashreporter is enabled or disabled) due to linux_syscall_support.h being missing. This PR pulls the required files directly from esr52.

Tested and confirmed that Linux build compiles successfully using GCC 7 on CentOS 7 with this patch.


wolfbeast:
Where did this go missing?...

trav90:
Not sure. It seems to have been missing since the initial commit here (though it does exist in the möbius repo).

PR #50, opened 2018-03-05
Fix configure on Mac OS X

### LINKED ISSUE #49: Mac OS X: build fails at configure stage
Python interpreter errors with:
`0:15.01 ['UnsortedError: An attempt was made to add an unsorted sequence to a list. The incoming list is unsorted starting at element 1. We expected "background.png" but got "disk.icns"\n']`

adeshkp:
This resolves #49


SpockMan02:
I can confirm that this fixes it.

PR #94, opened 2018-04-07
Fix Value::isGCThing footgun, stop returning true for NullValue

trav90:
This PR fixes the easily reproducible crashes experienced when UXP is built using GCC 7. From Mozilla ([Bug 1325075](https://bugzilla.mozilla.org/show_bug.cgi?id=1325075)):

> Value::isGCThing and Value::isMarkable do almost the same thing, but isGCThing returns true for NullValue.
We could remove isGCThing/toGCThing and convert all consumers to isMarkable. After that we can move JSVAL_TAG_NULL before JSVAL_LOWER_INCL_TAG_OF_GCTHING_SET and probably rename isMarkable/isGCThing to isGCThing/toGCThing.

I am not familiar with JS code at all so requesting an additional review before this gets merged. I have done some general browsing (including to pages that reliably crashed GCC 7 builds in the past) for a little bit now and have not seen any issues or experienced any crashes yet.

### DIFF
--- a/dom/xhr/XMLHttpRequestWorker.cpp
+++ b/dom/xhr/XMLHttpRequestWorker.cpp
@@ -1148,8 +1148,8 @@ EventRunnable::PreDispatch(WorkerPrivate* /* unused */)
       } else {
         bool doClone = true;
         JS::Rooted<JS::Value> transferable(cx);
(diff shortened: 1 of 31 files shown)


khronosschoty:
I tested this out and it appears to work great. I'll report back if I notice anything funny.  But so far things are flawless from what I can tell.

PR #100, opened 2018-04-07
Added support of the appmenu for DevTools menuitems (optional)

### LINKED ISSUE #96: DevTools - Added support of the appmenu (optional)
Some preparation for `Pale Moon UXP`.

janekptacijarabaci:
This resolves #96 → You add the label `Component: Devtools`, please

Reverted:
https://hg.mozilla.org/mozilla-central/rev/836da2d40b19
\+ adding some code

---

I've created the new build (x32, Windows) - `Basilisk`/`Pale Moon UXP` - and tested.

### DIFF
--- a/devtools/client/framework/browser-menus.js
+++ b/devtools/client/framework/browser-menus.js
@@ -156,6 +156,12 @@ function createToolMenuElements(toolDefinition, doc) {
     });
   }
 
+  let appmenuitem = createMenuItem({
+    doc,
+    id: "appmenuitem_" + id,
+    label: toolDefinition.menuLabel || toolDefinition.label
+  });
+
   let menuitem = createMenuItem({
     doc,
     id: "menuitem_" + id,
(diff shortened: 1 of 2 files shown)


mattatobin:
What happens when there isn't an appmenu?

mattatobin:
Blocks #102

janekptacijarabaci:
> What happens when there isn't an appmenu?

IMHO: The menu should be unified. And some users uses this menu - only:
https://github.com/MoonchildProductions/Pale-Moon/issues/1588

> Also, I want to keep devtools optional.. Keep that in mind.

IMHO: This removal (to an extension) does not change anything. They're still the same files, I just added the code for support the app menu. And if it does not exist, nothing will happen.

mattatobin:
Works for me.. Proceed.

Also, stop with the "IMHO" please just state it.. Thanks! :)

PR #101, opened 2018-04-07
CustomizableUI.jsm can't be used for Pale Moon - setting conditions

### LINKED ISSUE #97: DevTools - CustomizableUI can't be used for Pale Moon
Some preparation for `Pale Moon UXP`.

janekptacijarabaci:
This resolves #97 → You add the label `Component: Devtools`, please

Or is there any better solution for this?

---

I've created the new build (x32, Windows) - `Basilisk`/`Pale Moon UXP` - and tested.

### DIFF
--- a/devtools/client/framework/devtools-browser.js
+++ b/devtools/client/framework/devtools-browser.js
@@ -29,7 +29,7 @@ loader.lazyRequireGetter(this, "BrowserMenus", "devtools/client/framework/browse
 
 loader.lazyImporter(this, "AppConstants", "resource://gre/modules/AppConstants.jsm");
 #ifdef MC_BASILISK
-  loader.lazyImporter(this, "CustomizableUI", "resource:///modules/CustomizableUI.jsm");
+loader.lazyImporter(this, "CustomizableUI", "resource:///modules/CustomizableUI.jsm");
 #endif
 
 const {LocalizationHelper} = require("devtools/shared/l10n");
--- a/devtools/client/framework/devtools-browser.js
+++ b/devtools/client/framework/devtools-browser.js
@@ -27,8 +27,11 @@ loader.lazyRequireGetter(this, "DebuggerServer", "devtools/server/main", true);
 loader.lazyRequireGetter(this, "DebuggerClient", "devtools/shared/client/main", true);
 loader.lazyRequireGetter(this, "BrowserMenus", "devtools/client/framework/browser-menus");
 
-loader.lazyImporter(this, "CustomizableUI", "resource:///modules/CustomizableUI.jsm");
 loader.lazyImporter(this, "AppConstants", "resource://gre/modules/AppConstants.jsm");
+
+if (AppConstants.MOZ_APP_NAME.toLowerCase() != "palemoon") {
+  loader.lazyImporter(this, "CustomizableUI", "resource:///modules/CustomizableUI.jsm");
+}
 
 const {LocalizationHelper} = require("devtools/shared/l10n");
 const L10N = new LocalizationHelper("devtools/client/locales/toolbox.properties");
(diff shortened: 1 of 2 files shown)


devtools/client/framework/devtools-browser.js
```diff
@@ -27,9 +27,12 @@ loader.lazyRequireGetter(this, "DebuggerServer", "devtools/server/main", true);
 loader.lazyRequireGetter(this, "DebuggerClient", "devtools/shared/client/main", true);
 loader.lazyRequireGetter(this, "BrowserMenus", "devtools/client/framework/browser-menus");
 
-loader.lazyImporter(this, "CustomizableUI", "resource:///modules/CustomizableUI.jsm");
 loader.lazyImporter(this, "AppConstants", "resource://gre/modules/AppConstants.jsm");
 
+if (AppConstants.MOZ_APP_NAME.toLowerCase() != "palemoon") {
```
mattatobin: Please do NOT use AppConstants.. That will go away...

Do like I did in 8f6d3dab81c7f8f97ef197e26ab9439b09735b8f

You will need to locate the final target path for this file in dist and then take it out of the crappy mozbuild function magic and put the file into preprocessing..

You would then ifdef it with the conditional `MC_BASILISK`.

```
#ifdef MC_BASILISK
  loader.lazyImporter(this, "CustomizableUI", "resource:///modules/CustomizableUI.jsm");
#endif
```

devtools/client/framework/devtools-browser.js
```diff
@@ -295,6 +298,9 @@ var gDevToolsBrowser = exports.gDevToolsBrowser = {
    * Install Developer widget
    */
   installDeveloperWidget: function () {
+    if (typeof CustomizableUI === "undefined") {
```
mattatobin: In these conditionals...

#ifdef MC_BASILISK
  //all the code
#else 
  return whatever;
#endif

devtools/client/framework/devtools-browser.js
```diff
@@ -350,7 +356,7 @@ var gDevToolsBrowser = exports.gDevToolsBrowser = {
    */
   // Used by itself
   installWebIDEWidget: function () {
-    if (this.isWebIDEWidgetInstalled()) {
+    if ((typeof CustomizableUI === "undefined") || this.isWebIDEWidgetInstalled()) {
```
mattatobin: ditto

devtools/client/framework/devtools-browser.js
```diff
@@ -374,6 +380,9 @@ var gDevToolsBrowser = exports.gDevToolsBrowser = {
   },
 
   isWebIDEWidgetInstalled: function () {
+    if (typeof CustomizableUI === "undefined") {
```
mattatobin: ditto

devtools/client/framework/devtools-browser.js
```diff
@@ -387,6 +396,9 @@ var gDevToolsBrowser = exports.gDevToolsBrowser = {
    * Uninstall WebIDE widget
    */
   uninstallWebIDEWidget: function () {
+    if (typeof CustomizableUI === "undefined") {
```
mattatobin: ditto

devtools/client/framework/devtools-browser.js
```diff
@@ -398,6 +410,9 @@ var gDevToolsBrowser = exports.gDevToolsBrowser = {
    */
    // Used by webide.js
   moveWebIDEWidgetInNavbar: function () {
+    if (typeof CustomizableUI === "undefined") {
```
mattatobin: ditto

mattatobin:
Blocks #102

janekptacijarabaci:
If this is used:

```
#ifndef MC_PALEMOON
  loader.lazyImporter(this, "CustomizableUI", "resource:///modules/CustomizableUI.jsm");
#endif
```

and

```
#ifdef MC_PALEMOON
  return;
#endif
```

...it is likely that other frontends will use it (`CustomizableUI`).

Is it better than the original one (`#ifdef MC_BASILISK`)?

mattatobin:
Then you end up with ureachable code warnings being emited and there is no reason australis code should be there.. Preprocessing it out as i indicated is the prefered solution..

wolfbeast:
CustomizableUI is Australis-specific, so only in use in Basilisk on the platform.

PR #99, opened 2018-04-07
Remove unused variables (DevTools) in browser.js

janekptacijarabaci:
This resolves #95 → You add the label `Component: Devtools`, please

Bug(s):
https://bugzilla.mozilla.org/show_bug.cgi?id=1250832

---

I've created the new build (x32, Windows) and tested.

PR #106, opened 2018-04-08
[DevTools] Follow up: A function "Inspect Element" (nsContextMenu.js)

janekptacijarabaci:
Follow up: #99

`gDevTools` is used in `nsContextMenu.js`


mattatobin:
gDevTools is gonna have to be made conditional.. It will fail to load without devtools won't it?

Issue #57, opened 2018-03-12
Port Pale Moon to UXP


mattatobin:
This is the Meta bug for porting `Pale Moon` to `UXP`.

----

- [x] Make `Pale Moon` Build
    - [x] Basic `moz.build` changes
    - [x] Basic `c++` changes
- [x] Changes to front-end code to account for back-end changes
    - [x] Initial loading of browser window
    - [x] Developer Tools (Issue #102)
    - [x] Fullscreen (Issue #270) - _No longer blocking unstable_
    - [x] Findbar
    - [x] Places (History, Bookmarks, and Downloads)
      - [x] Library Window
      - [x] Downloads Panel
      - [x] Boomarks Sidebar
      - [x] History Sidebar
    - [x] Preferences Window - _Does not block unstable_
    - [x] Doorhangers
        - [x] Odd checkbox element present
    - [x] Sync - _Does not block unstable_
    - [x] Login/Password Manager
    - [x] New Tab Page (about:newtab) - _Does not block unstable_
    - [x] Shell Service
    - [x] URL bar
    - [x] Session Store
    - [x] Feeds Service
    - [x] **_More as they are determined_**
- [ ] Add-ons
    - [x] Extensions
      - [x] Bootstrap
      - [x] Jetpack (Issue #120)
    - [x] Default Theme - _No longer blocking unstable_
    - [ ] Language Packs - _Does not block unstable_
- [x] Packaging
    - [x] `package-manifest.in` / `removed-files.in`
    - [x] Windows Installer
- [x] Default Preferences Audit - _Does not block unstable_
    - [x] Application Preferences
    - [x] Branding Preferences

wolfbeast:
Sync has been fixed, ticked it off.

wolfbeast:
Ticked off several more things that are fixed. Almost there ;)

wolfbeast:
Everything but the language packs is addressed -- closing.

JustOff:
In fact, language packs are almost ready, but we need to freeze the tree against string changes so that the translation can be finalized before the release.

@wolfbeast, please don't forget to create a corresponding task in [pale-moon-localization](https://github.com/MoonchildProductions/pale-moon-localization) when you decide that the time has come.

JustOff:
Roger that.

PR #124, opened 2018-04-12
about:plugins - fix "addMessageListener is not a function" (e10s off - revert bug 1068087 - partially)

janekptacijarabaci:
Tag: #57 

__Steps to reproduce__
Go to: `about:plugins`
Throws an error: `addMessageListener is not a function`

e10s off - revert [bug 1068087](https://bugzilla.mozilla.org/show_bug.cgi?id=1068087) - partially (frontend only).

---

### DIFF
--- a/toolkit/content/plugins.html
+++ b/toolkit/content/plugins.html
@@ -10,6 +10,7 @@
   "use strict";
 
   Components.utils.import("resource://gre/modules/Services.jsm");
+  Components.utils.import("resource://gre/modules/AddonManager.jsm");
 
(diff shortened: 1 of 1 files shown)


mattatobin:
Does this remain working for both `TychoAM` and `WebExAM`?

janekptacijarabaci:
It works for Basilisk and Pale Moon (if that's enough). But for Basilisk with e10s off only.

I'll adjust it...

Thanks.

mattatobin:
No.. we have zero intention of supporting e10s in Basilisk.. it is fine.

PR #207, opened 2018-04-19
 moebius#73: DOM - window.requestIdleCallback - improvements

janekptacijarabaci:
Tag https://github.com/MoonchildProductions/moebius/pull/73

__+ depends on:__
- https://bugzilla.mozilla.org/show_bug.cgi?id=1336229
  - Don't dispatch canceled IdleRequestExecutors
- https://bugzilla.mozilla.org/show_bug.cgi?id=1334904
  - Add test for when rIC timeouts doesn't need timeout 
- https://bugzilla.mozilla.org/show_bug.cgi?id=1337814
  - Add test for cancelling currently executing rIC callback
  - Remove rIC callback from pending callbacks before running it

---

`about:config`
`dom.requestIdleCallback.enabled` = `true`

__An example:__
https://developer.mozilla.org/en-US/docs/Web/API/Background_Tasks_API#Example

---


wolfbeast:
Main implementation/improvements for #209

Issue #121, opened 2018-04-12
Front-end issues with Pale Moon application


wolfbeast:
A collection of front-end issues with the Pale Moon application on UXP.

- [x] Browser console: source links to chrome:// URIs load content page instead
- [x] `nsFaviconService::setAndFetchFaviconForPage()` deprecated, please use `nsFaviconService::setAndFetchFaviconForPage(..., [optional aLoadingPrincipal])` tabbrowser.xml:776:14
- [x] `nsIOService::NewChannelFromURI()` deprecated, please use `nsIOService::NewChannelFromURI2()` WindowsPreviewPerTab.jsm:77:16
- [x] `nsICookieManager2.getCookiesFromHost()` is changed. Update your code and pass the correct originAttributes. Read more on MDN: https://developer.mozilla.org/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsICookieManager2 SessionStore.jsm:2405:17
- [x] Warning: Property contained reference to invalid variable.  Error in parsing value for ‘background-color’.  Falling back to ‘initial’.
Source File: chrome://devtools/skin/webconsole.css
Line: 406, Column: 9402
Source Code:
` var(--theme-tab-toolbar-background)`
- [x] Warning: Property contained reference to invalid variable.  Error in parsing value for ‘border-top-width’.  Falling back to ‘initial’.
Source File: chrome://devtools/skin/webconsole.css
Line: 407, Column: 9459
Source Code:
` 1px solid var(--theme-splitter-color)`
- [x] Warning: Property contained reference to invalid variable.  Error in parsing value for ‘border-top-width’.  Falling back to ‘initial’.
Source File: chrome://devtools/skin/widgets.css
Line: 84, Column: 2850
Source Code:
` var(--devtools-splitter-top-width)`
- [x] Warning: Property contained reference to invalid variable.  Error in parsing value for ‘border-bottom-width’.  Falling back to ‘initial’.
Source File: chrome://devtools/skin/widgets.css
Line: 85, Column: 2908
Source Code:
` var(--devtools-splitter-bottom-width)`
- [x] Places: Error: SyntaxError: non-generator method definitions may not contain yield
Source File: resource:///modules/MigrationUtils.jsm
Line: 510, Column: 8
Source Code:
`        yield migrator; `
- [x] Places: Error: ReferenceError: can't access lexical declaration `ContentArea' before initialization
Source File: chrome://browser/content/places/downloadsViewOverlay.xul
Line: 21
[...]

PR #215, opened 2018-04-20
[PALEMOON] Reimplemented "Time Zone Information"

janekptacijarabaci:
Tags:
https://github.com/MoonchildProductions/Pale-Moon/issues/593
https://github.com/MoonchildProductions/Pale-Moon/pull/599
Issue #121

---

### DIFF
--- a/application/palemoon/app/profile/palemoon.js
+++ b/application/palemoon/app/profile/palemoon.js
@@ -80,6 +80,14 @@ pref("browser.slowstartup.help.url", "http://www.palemoon.org/support/slowstartu
 // on a network error page (e.g. cert error)
 pref("browser.escape_to_blank", false);
 
+#ifdef XP_WIN
+// Save internet zone information on downloaded files:
+// 0 => Never
+// 1 => Always
+// 2 => Use system setting
+pref("browser.download.saveZoneInformation", 2);
+#endif
+
 // The minimum delay in seconds for the timer to fire.
 // default=2 minutes
 pref("app.update.timerMinimumDelay", 120);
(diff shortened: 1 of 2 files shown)


wolfbeast:
This has nothing to do with time zones...

wolfbeast:
This also has nothing to do with front end integration (121)... 
Can you please be more careful how you name/link/describe issues?

wolfbeast:
Tobin already merged it I see. well, that sucks. Sorry but I'll back this out and redo it so the commit history doesn't have confusing info.

mattatobin:
I looked ar the code not the description because i can never trust it.. sorry.

janekptacijarabaci:
> This also has nothing to do with front end integration (121)...

AFAIK: It is in `Preferences` - `General`:
preferences

But of course it's up to you.

mattatobin:
The patch was good but your commit message was incorrect also this should eventually get implimented for basilisk too so the conditional was ok for now.

wolfbeast:
Actually the `#ifdef MC_PALEMOON` is wrong. this is toolkit/platform stuff.

janekptacijarabaci:
> commit message was incorrect

Yes, unfortunately, it was my fault...

> Actually the #ifdef MC_PALEMOON is wrong

So using `#ifdef MC_BASILISK` only?

wolfbeast:
> So using `#ifdef MC_BASILISK` only?

No! It should apply to any application that downloads files on Windows.

mattatobin:
Is the pref the only bit needed by the code then?

wolfbeast:
All it ever did was make it prefable instead of always applying based on system.

PR #222, opened 2018-04-20
 Force use of level 1 Liftetime DSE optimization with modern GCC 

### LINKED ISSUE #14: Attempt to work around GCC 6's overzealous optimizations
`nsIPresShell::AllocateFrame` and `::AllocateByObjectID` implement a pool allocator that zeroes out returned memory. This allocator is in place so that we don't have to worry about uninitialized variables in related objects.

Unfortunately this falls foul of GGC 6's `lifetime-dead-store-elimination` optimisations.  This detects
the `memset(..)` as occurring _before_ the lifetime of the allocated object begins, and so removes the call in its infinite wisdom. This in turn causes `nsFrame` and many child classes to have uninitialized fields after construction (normally zeroed out), which leads to multiple segfaults in short order.

Removing the use of this zeroing allocator is an option, but risky, because then we'd have to go through all the consumers of it and manually initialize the variables.

trav90:
Just what it says in the tin. Resolves #14 using option 2 discussed in that issue.

### DIFF
--- a/build/autoconf/compiler-opts.m4
+++ b/build/autoconf/compiler-opts.m4
@@ -183,16 +183,15 @@ if test "$GNU_CC"; then
         case "$CC_VERSION" in
         4.*)
             ;;
+        5.*)
+            ;;
         *)
             # Lifetime Dead Store Elimination level 2 (default in GCC6+) breaks Gecko.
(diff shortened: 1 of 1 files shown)


khronosschoty:
Very nice! Cheers!

PR #269, opened 2018-04-26
Improve some button labels in preferences

### LINKED ISSUE #253: Improve some button labels in preferences
Options > "Privacy" tab > "remove individual cookies" link

Here are the current Pale Moon messages:
* when there is no search filter:
Remove Cookie - Remove All Cookies
* when there is a search filter:
Remove Cookie[s] - Remove All Cookies

I found it confusing ("a search filter is applied, if I click "Remove All Cookies", will it remove only the shown cookies or all the cookies?")

Then I gave a run with Firefox, and noticed they are using much less ambiguous messages:
* when there is no search filter:
Remove Selected - Remove All
* when there is a search filter:
Remove Selected - Remove All Shown

I think these clearer messages should be ported to Pale Moon.

https://bugzilla.mozilla.org/show_bug.cgi?id=1338036 - RESOLVED FIXED in Firefox 54

adeshkp:
This resolves #253

### DIFF
--- a/application/palemoon/components/preferences/cookies.js
+++ b/application/palemoon/components/preferences/cookies.js
@@ -5,6 +5,8 @@
 
 const nsICookie = Components.interfaces.nsICookie;
 
+Components.utils.import("resource://gre/modules/PluralForm.jsm");
+
 var gCookiesWindow = {
   _cm               : Components.classes["@mozilla.org/cookiemanager;1"]
                                 .getService(Components.interfaces.nsICookieManager),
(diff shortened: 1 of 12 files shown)


vlakoff:
That's an impressive work, thank you very much!

JustOff:
Could someone explain me why `removeSelectedCookies.label` uses plurals, although the value does not contain any number field, i.e `#1`?

It seems quite unnecessary and I would like to remove it to simplify localization.

wolfbeast:
"Remove selected cookies" can pertain to 1 cookie or multiple cookies (depending on whether you have selected 1 cokie or multiple). The wording is different depending. (Remove selected cookie/Remove selected cookies). even if "cookie/cookies" is omitted, it can need different word formes for "selected" in other languages.

JustOff:
Ok, thanks.

PR #271, opened 2018-04-26
[PALEMOON] [frontend vs backend] Move tab to a new window does not work properly

janekptacijarabaci:
Tag #121

Throws an error:
```
TypeError: PopupNotifications._swapBrowserNotifications is not a function
browser.xml:1349:13
```

Bug(s):
https://bugzilla.mozilla.org/show_bug.cgi?id=723951

---


mattatobin:
It would be nice if we could re-align with the toolkit version of PopupNotifications.jsm

Issue #286, opened 2018-04-28
Incorrect osx include filepaths


SpockMan02:
`/browser/themes/osx/shared.inc` line 1:
`%include ../../../../toolkit/themes/osx/global/shared.inc`
This causes build failure and should be `../../../toolkit/themes/osx/global/shared.inc`
After solving #285 and this I have been able to successfully build Serpent.

SpockMan02:
That fixes it, but it sounds like it'll need to be reverted when Basilisk is put in /application. Feel free to close or keep open as a reminder.

mattatobin:
Yes.

mattatobin:
Blocks #303

SpockMan02:
Except that we've got a similar problem in another place now.
`/application/basilisk/base/content/baseMenuOverlay.xul` line 32
looks for `../../../toolkit/content/macWindowMenu.inc`
FILE_NOT_FOUND

wolfbeast:
That include has been fixed now. Anything else? :)

SpockMan02:
Yep, gimme a sec, there's one more that I just fixed locally.

wolfbeast:
That's it, then?

SpockMan02:
And with that, it builds for me. :)

PR #287, opened 2018-04-28
Issue #286 - Incorrect filepath

SpockMan02:


mattatobin:
I set it in prep for moving Basilisk so it would already be done but moving it kept getting pushes back so r+

Issue #55, opened 2018-03-12
DOM - implement animationcancel event


wolfbeast:
Cloned from https://github.com/MoonchildProductions/moebius/issues/85

https://drafts.csswg.org/css-animations/#eventdef-animationevent-animationcancel

> animationcancel

> The animationcancel event occurs when the animation stops running in a way that does not
fire an animationend event, such as a change in the animation-name that removes the animation, or the animating element or one of its ancestors becoming display:none.

> The elapsedTime member for this event indicates the number of seconds that had elapsed since the beginning of the animation at the moment when the animation was canceled. This excludes any time where the animation was paused. If the animation had a negative animation-delay, the beginning of the animation is the moment equal to the absolute value of animation-delay seconds prior to when the animation was actually triggered. Alternatively, if the animation had a positive animation-delay and the event is fired before the animation’s delay has expired, the elapsedTime will be zero.

> Bubbles: Yes
> Cancelable: No
> Context Info: animationName, elapsedTime, pseudoElement 

Not directly portable from Möbius, probably needs some prereq code changes.

janekptacijarabaci:
Targeting bugs are: [1202333](https://bugzilla.mozilla.org/show_bug.cgi?id=1202333) and [1264125](https://bugzilla.mozilla.org/show_bug.cgi?id=1264125)

janekptacijarabaci:
__Native in moebius:__
[Bug 1324985: Firing the `ontransitionrun` and `ontransitionstart`](https://bugzilla.mozilla.org/show_bug.cgi?id=1324985)
\- revert this (implemented for tree 52 and 52.x only)

IMHO, this is closely related to https://github.com/MoonchildProductions/moebius/issues/85:

[#c0](https://bugzilla.mozilla.org/show_bug.cgi?id=1324985#c0):
> I landed transitionrun and transitionstart implementation on bug 1287983(Aurora 52), then I landed transitioncancel on bug 1264125(Nightly 53).

__https://drafts.csswg.org/css-transitions-1/#transitionrun__

> transitionrun
> - The transitionrun event occurs when a transition is created (i.e., when it is added to the set of running transitions).
>
> A negative transition-delay will cause the event to fire with an elapsedTime equal to the absolute value of the delay capped to the transition-duration of the animation. That is, the elapsed time is equal to min(max(-transition-delay, 0), transition-duration).
>
> Bubbles: Yes
> Cancelable: No
[...]

PR #296, opened 2018-04-29
DOM - implement animationcancel event

janekptacijarabaci:
Tag #55 

__Native in moebius:__
[Bug 1324985: Firing the `transitionrun` and `transitionstart`](https://bugzilla.mozilla.org/show_bug.cgi?id=1324985)
\- revert this (implemented for tree 52 and 52.x only)

IMHO, this is closely related to https://github.com/MoonchildProductions/moebius/issues/85:

[#c0](https://bugzilla.mozilla.org/show_bug.cgi?id=1324985#c0):
> I landed transitionrun and transitionstart implementation on bug 1287983(Aurora 52), then I landed transitioncancel on bug 1264125(Nightly 53).

__https://drafts.csswg.org/css-transitions-1/#transitionrun__

> transitionrun
> - The transitionrun event occurs when a transition is created (i.e., when it is added to the set of running transitions).
>
> A negative transition-delay will cause the event to fire with an elapsedTime equal to the absolute value of the delay capped to the transition-duration of the animation. That is, the elapsed time is equal to min(max(-transition-delay, 0), transition-duration).
>
> Bubbles: Yes
> Cancelable: No
> Context Info: propertyName, elapsedTime, pseudoElement 

> transitionstart
> - The transitionstart event occurs when a transition’s delay phase ends.
>
> The value of elapsedTime for transitionstart events is the same as the value used for transitionrun events.
>
> Bubbles: Yes
> Cancelable: No
> Context Info: propertyName, elapsedTime, pseudoElement 

__An example:__
https://developer.mozilla.org/en-US/docs/Web/Events/transitionrun

---

__For `animationcancel` event:__

Depends on:
[Bug 1202333: AnimationEvent elapsedTime should reflect playbackRate](https://bugzilla.mozilla.org/show_bug.cgi?id=1202333)
[Bug 1264125: Fire transitioncancel event when a transition is canceled](https://bugzilla.mozilla.org/show_bug.cgi?id=1264125)

And implemented:
https://github.com/MoonchildProductions/moebius/pull/89

__An example:__
https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onanimationcancel

---

__A new feature in the core - `Verification Needed`, please.__


wolfbeast:
New features should have an issue, not just a PR.

wolfbeast:
Resolves #55

PR #300, opened 2018-04-30
[PALEMOON] Fix a build warning for autocomplete.css

janekptacijarabaci:
Build - warnings:
```
[drive]:\[path]\application\palemoon\base\content/autocomplete.css:
WARNING: no preprocessor directives found
[drive]:\[path]\application\palemoon\base\content/autocomplete.xml:
WARNING: no preprocessor directives found
```


mattatobin:
r+ on the css file
r- on the xml file because AppConstants are in effect.. Non-fatal preprocessor warning will become n/a once I add some preprocessor directives

janekptacijarabaci:
> How about we just leave it alone until I de-AppConstants things

OK, I'm leaving it for you...

PR #231, opened 2018-04-23
moebius#223, #224, #226, #230: DOM - consider blocking top level window data: URIs

janekptacijarabaci:
## Implemented:

__From moebius:__
https://github.com/MoonchildProductions/moebius/pull/223
https://github.com/MoonchildProductions/moebius/pull/224
https://github.com/MoonchildProductions/moebius/pull/226
https://github.com/MoonchildProductions/moebius/pull/230

~~https://github.com/MoonchildProductions/moebius/pull/231~~
\- it will be added later

__Depends on:__
- Native in moebius
  - [Bug 1329288](https://bugzilla.mozilla.org/show_bug.cgi?id=1329288)
    - Part 1/3: It's implemented
    - Part 2/3: Test ContentPolicy blocks opening a new window
    - Part 3/3: Allow content policy consumers to identify contentPolicy checks from docshell
  - [Bug 1329570](https://bugzilla.mozilla.org/show_bug.cgi?id=1329570)
    - Restore missing error check in mozilla::dom::Location::GetPathname
  - [Bug 1323683](https://bugzilla.mozilla.org/show_bug.cgi?id=1323683)
    - Fold nsIURIWithQuery into nsIURI
- [moebius#187: DOM - nsIContentPolicy - context (document)](https://github.com/MoonchildProductions/moebius/pull/187)

---

I've created the new build (x32, Windows) - `Basilisk` / `Pale Moon UXP` - and tested.

PR #311, opened 2018-05-01
Use asyncOpen2() for docshell loads

### LINKED ISSUE #310: XPCOM - "shouldLoad()" is not running
__Steps to reproduce:__

Install `Greasemonkey for Pale Moon`
Drag & Drop an user script (*.user.js) on the main window

__Actual results:__
The file is only displayed.

__Expected results:__
`shouldLoad()` should be started
(https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIContentPolicy)
and an installation window is displayed.

janekptacijarabaci:
This resolves #310 

__Follow up:__
#231

__Bug(s):__
https://bugzilla.mozilla.org/show_bug.cgi?id=1182569
\+ depends on
(native in moebius)

---

For `Pale Moon UXP` - `triggeringPrincipal` is not used.

---

It is a bigger change - please add the label `Verification Needed`.

### DIFF
--- a/browser/base/content/browser.js
+++ b/browser/base/content/browser.js
@@ -2779,24 +2779,6 @@ var BrowserOnClick = {
     }
   },
 
-  handleEvent: function (event) {
-    if (!event.isTrusted || // Don't trust synthetic events
-        event.button == 2) {
-      return;
-    }
-
-    let originalTarget = event.originalTarget;
-    let ownerDoc = originalTarget.ownerDocument;
-    if (!ownerDoc) {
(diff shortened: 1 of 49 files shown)


JustOff:
As far as I see, here is no changes in Pale Moon application code, does it mean that updated docshell API is just backward compatible?

JustOff:
@janekptacijarabaci, it turns out I was not in vain worried about this :sunglasses:

Issue #120, opened 2018-04-11
Make Add-on SDK (Jetpack) work on Pale Moon


JustOff:
This is the Meta bug for making `Add-on SDK` (Jetpack) work on `Pale Moon`.

---

- [x] Add `Pale Moon` as a target engine to `sdk/system/xul-app.jsm`
- [x] Add `Pale Moon` to `modules/system/Startup.js` (bugs [851426](https://bugzilla.mozilla.org/show_bug.cgi?id=851426) and [1042239](https://bugzilla.mozilla.org/show_bug.cgi?id=1042239))
- [x] Update `module.metadata` with the `Pale Moon` engine where appropriate in `sdk/*`
- [x] Add the `PMkit` shim for `sdk/ui/button` ([PM27#808](https://github.com/MoonchildProductions/Pale-Moon/pull/808), [PM27#818](https://github.com/MoonchildProductions/Pale-Moon/pull/818), [PM27#918](https://github.com/MoonchildProductions/Pale-Moon/pull/918))
- [x] Allow access to `sdk/ui/buttons` via `require('sdk/ui')` ([PM27#811](https://github.com/MoonchildProductions/Pale-Moon/pull/811))
- [x] Deal with `sdk/tabs`
  - [x] ~Fix memory leak ([PM27#776](https://github.com/MoonchildProductions/Pale-Moon/pull/776))~ already fixed in UXP
  - [x] ~Investigate the possibility to de-e10s ([PM27#785](https://github.com/MoonchildProductions/Pale-Moon/pull/785))~ see #331
- [x] Deal with `sdk/context-menu` and `sdk/context-menu@2`
  - [x] Provide gContextMenuContentData ~([PM27#778](https://github.com/MoonchildProductions/Pale-Moon/pull/778), [PM27#780](https://github.com/MoonchildProductions/Pale-Moon/pull/780))~ (steal from Basilisk)
  - [x] ~Fix TypeError in messageManager ([PM27#783](https://github.com/MoonchildProductions/Pale-Moon/pull/783))~ already fixed in UXP
  - [x] ~Investigate the possibility to de-e10s ([PM27#784](https://github.com/MoonchildProductions/Pale-Moon/pull/784))~ see #331
  - [x] ~Remove `sdk/context-menu@2` ([PM27#787](https://github.com/MoonchildProductions/Pale-Moon/pull/787))~ see #331
[...]

PR #330, opened 2018-05-04
[PMkit] Fix Empty string passed to getElementById() warning in buttons.js

JustOff:
Just what it says in the tin.

Tag #120.

### DIFF
--- a/toolkit/jetpack/sdk/ui/buttons.js
+++ b/toolkit/jetpack/sdk/ui/buttons.js
@@ -50,7 +50,7 @@ function insertButton(aWindow, id, onBuild) {
   let toolbar = toolbarId != "" && doc.getElementById(toolbarId);
 
   if (toolbar) {
-    let nextItem = doc.getElementById(nextItemId);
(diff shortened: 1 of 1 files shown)


toolkit/jetpack/sdk/ui/buttons.js
```diff
@@ -50,7 +50,7 @@ function insertButton(aWindow, id, onBuild) {
   let toolbar = toolbarId != "" && doc.getElementById(toolbarId);
 
   if (toolbar) {
-    let nextItem = doc.getElementById(nextItemId);
+    let nextItem = nextItemId != "" && doc.getElementById(nextItemId);
```
adeshkp: Just use `nextItemId`, i.e.
`let nextItem = nextItemId && doc.getElementById(nextItemId);`

JustOff: I prefer to use explicit comparisons for easier understanding. JS compiler is smart enough to skip the possible overhead.

Issue #270, opened 2018-04-26
[PALEMOON] FullScreen - this feature does not work properly


janekptacijarabaci:
Tag #121 

F11 key make window full screen but the upper bar remains visible (and the CSS is completely wrong here).

There are many changes. Maybe it would be best to try adding e.g. (without `browser-fullScreen.js`):
[fullscr-toggler element needs to be hidden in DOM fullscreen](https://hg.mozilla.org/mozilla-central/rev/cfbe5fc74ea2)
[Avoid exiting fullscreen mode when exit from DOM fullscreen](https://hg.mozilla.org/mozilla-central/rev/4ac24fe93a88)
[Prevent collapsing toolbar on fullscreen mode on OS X](https://hg.mozilla.org/mozilla-central/rev/d11b9d29d909)
[Skip changing fullscreen window controls when using OS X Lion native fullscreen](https://hg.mozilla.org/mozilla-central/rev/1bc7f9b2e95c)
[Skip hiding toolbars for fullscreen mode on OS X Lion](https://hg.mozilla.org/mozilla-central/rev/9a2ffd3b6381)
[Simplify fullscreen code](https://hg.mozilla.org/mozilla-central/rev/1f7db5d87e80)
[Restore the position the fullscreen toolbox collapsed flag is set](https://hg.mozilla.org/mozilla-central/rev/f021a650dd49)
[Work around the inDOMFullscreen attribute being removed too late by checking inFullscreen too](https://hg.mozilla.org/mozilla-central/rev/521e4737222b)
[Simplify gIdentityHandler and show connection type in the fullscreen notification](https://hg.mozilla.org/mozilla-central/rev/15fd1ffcc1be)
[Retry hiding toolbox when some of safe-to-collapse conditions change](https://hg.mozilla.org/mozilla-central/rev/f944a86418a9)
[Remove pointerlock permission requirement](https://hg.mozilla.org/mozilla-central/rev/06241d818e99)
[Follow up to fix localisation](https://hg.mozilla.org/mozilla-central/rev/6648e2942cf9)
[Remove UI leftovers from the pointerLock removal](https://hg.mozilla.org/mozilla-central/rev/863ce9e133e8)

and then add the file `browser-fullScreenAndPointerLock.js` (without some e10s work) from `/browser` (Basilisk).

But I don't know...

JustOff:
I guess https://bugzilla.mozilla.org/show_bug.cgi?id=1161802 is the key.

[...]

PR #334, opened 2018-05-06
[PALEMOON] Replace the fullscreen permission dialog with a warning and update the permission management

JustOff:
This changes the following:

* Update fullscreen activation event to match UXP
* Replace fullscreen permission dialog with a warning
* Remove fullscreen and pointer-lock from permission managing (not available in UXP)

Based on:

* https://bugzilla.mozilla.org/show_bug.cgi?id=1053413
* https://bugzilla.mozilla.org/show_bug.cgi?id=1160017

Since this removes `full-screen-api.approval-required`, I provided a way to achieve the same effect by setting `full-screen-api.warning.timeout=0` (improvement over Mozilla).

Tag #270.


mattatobin:
Blocks #57

Issue #102, opened 2018-04-08
Make DevTools work on Pale Moon


mattatobin:
This is a meta bug for all pull requests to initially get devtools working on Pale Moon.

@janekptacijarabaci will link future devtools pull requests to this one until DevTools is fully operational in Pale Moon instead of creating dozens of separate issues..

Blocks #57

mattatobin:
Blocks #57

mattatobin:
As near as I can tell the DevTools them selves seem fully functional.. What is your assessment of the devtools status, @janekptacijarabaci?

janekptacijarabaci:
Most features are functional (but I haven't tested it in detail).
Currently (e.g.): I see some bugs with Developer Toolbar. But that's a special category...

mattatobin:
Understood. Keep up the good work.

mattatobin:
Can we get a status report.. Is this finished in the scope of Pale Moon regardless if DevTools them selves need further enhancements.. It is a blocker after all...

mattatobin:
@janekptacijarabaci

mattatobin:
Thanks!

JustOff:
Sorry, but I just can't figure out how to enable the logging of request bodies in new devtools:
![2018-05-06_18-41-57](https://user-images.githubusercontent.com/340855/39675061-26ec103e-5145-11e8-83ab-51cb5a981018.png)
@janekptacijarabaci, could you give me a clue?

janekptacijarabaci:
@JustOff 
See:
[bug 1064458](https://bugzilla.mozilla.org/show_bug.cgi?id=1064458) Remove 'Log request and response bodies' preference
and
https://hg.mozilla.org/mozilla-central/rev/8d816bd77d80#l10.154

In `Browser Console` - now the result will always be: `Response body was not stored.`
In `Web Console` - there is everything.

JustOff:
@janekptacijarabaci, thanks for the info!

Too bad, it was a pretty stupid decision :-1:

wolfbeast:
No. the browser console is the console for the browser code, not content. There won't ever ber request and response bodies there because you're not debugging CONTENT there. If you are, then you're using the wrong tool.

JustOff:
I'm probably missing something, but what is the right tool, say, to inspect response body for extensions or blocklist update requests? I don't see anything suitable for this task in new devtools.

PR #337, opened 2018-05-07
[regression] DevTools - Web and Browser console - Added "Log request and response bodies" preference

janekptacijarabaci:
Tag #102 

IMHO: Debugging add-ons is an important thing.

__I suggest putting it into preferences:__
`devtools.webconsole.filter.saveBodies` (default = `true`)
`devtools.browserconsole.filter.saveBodies` (default = `false`)

__Implemented:__
- To simplify switches: the context menu isn't used
- `Log Request and Response Bodies` is a separate switch - after clicking on the menu category `Net` does not change

---

But you add the label `Verification Needed`, please.

### DIFF
--- a/devtools/client/locales/en-US/webConsole.dtd
+++ b/devtools/client/locales/en-US/webConsole.dtd
@@ -11,6 +11,11 @@
 <!ENTITY window.title "Web Console">
 <!ENTITY browserConsole.title "Browser Console">
 
+<!-- LOCALIZATION NOTE (saveBodies.label): You can see this string in
+   - the Browser Console context menu. -->
+<!ENTITY saveBodies.label     "Log Request and Response Bodies">
+<!ENTITY saveBodies.accesskey "L">
+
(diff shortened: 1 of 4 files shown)


wolfbeast:
... Where does this add the option, if not in the context menu?

janekptacijarabaci:
As the context menu is meant:
context-menu
not:
context-menu_main

wolfbeast:
I think you misunderstood my question.
So, if I understand correctly, it will be in the "Net" menu with this change, but not in the context menu? That's fine, if so. If I misunderstood then please explain what the result is.

janekptacijarabaci:
> So, if I understand correctly, it will be in the "Net" menu with this change, but not in the context menu?

Yes, that's the way it is.

JustOff:
This is exactly what I wanted, and it works perfectly!
Thank you @janekptacijarabaci :+1:

wolfbeast:
Okay! That works.

PR #339, opened 2018-05-07
DevTools - fix a comment in webConsole.dtd (Web Console and Browser Console)

janekptacijarabaci:
Issue #102 and #337


devtools/client/locales/en-US/webConsole.dtd
```diff
@@ -12,7 +12,7 @@
 <!ENTITY browserConsole.title "Browser Console">
 
 <!-- LOCALIZATION NOTE (saveBodies.label): You can see this string in
-   - the Browser Console context menu. -->
+   - the Web Console and Browser Console context menu. -->
```
wolfbeast: If you're correcting this, correct it properly.
It won't be in the context menu in the browser console -- since you didn't implement that.

PR #349, opened 2018-05-13
Fix for mac osx packaging

wicknix:


mattatobin:
I'd prefer it if you would redo this pull request cleanly and NOT delete the entire file.

PR #359, opened 2018-05-15
Revive the quickdial page (about:newtab)

JustOff:
Correction of the typo made [here](https://github.com/MoonchildProductions/UXP/commit/2569bca9308968c778f750c3db23c08d95aa7230) restores normal operation of the quickdial page.

Tag #57.


mattatobin:
You're kidding.

JustOff:
With this typo `Grid_renderGrid()` just silently died deep inside the callback chain, interrupting further initialization of the quickdial page. It took me almost an hour to track down the culprit, and only then did I realize I could start by checking the latest changes in the `about:newtab` code and get the result in a minute :upside_down_face:

PR #363, opened 2018-05-16
this patch fixes a mach build failure

khronosschoty:
The patch was found here https://bugzilla.mozilla.org/show_bug.cgi?id=1384062 and resolves a mach build failure I reported elsewhere

@mattatobin .

PR #366, opened 2018-05-16
Make SystemResourceMonitor.stop more resilient to errors

### LINKED ISSUE #110: Mach fails after completing build on some distros
Assertion Error in Mach on some Distros..

Workaround is to use `make -f client.mk` directly instead of `mach build`

alaviss:
Bug: [1384062](https://bugzilla.mozilla.org/show_bug.cgi?id=1384062)

Tags:
- MoonchildProductions/Pale-Moon#1414 (same issue occurred in UXP)
- #110 (described simillar issue)
- MoonchildProductions/Pale-Moon@d3e2e53ed55c82b6f68b0412edea7ea92ddd83e5 (the same patch in Pale Moon)

The fix as described by @wolfbeast:

> - Deal with the fact that on Linux, pipe isn't actually a pipe, meaning ends can die independently.
> - Establish a timeout so we don't hang forever if the child process has crashed.
> - Reduce the poll timeout to 0.1 since we don't have to wait that long anymore now the underlying issue has been fixed.

The main difference between this PR and #363 is that the commit message and authorship is preserved.

### DIFF
--- a/testing/mozbase/mozsystemmonitor/mozsystemmonitor/resourcemonitor.py
+++ b/testing/mozbase/mozsystemmonitor/mozsystemmonitor/resourcemonitor.py
@@ -289,26 +289,45 @@ class SystemResourceMonitor(object):
         assert self._running
         assert not self._stopped
 
-        self._pipe.send(('terminate',))
+        try:
+            self._pipe.send(('terminate',))
+        except Exception:
+            pass
         self._running = False
         self._stopped = True
 
         self.measurements = []
 
-        done = False
-
         # The child process will send each data sample over the pipe
         # as a separate data structure. When it has finished sending
         # samples, it sends a special "done" message to indicate it
         # is finished.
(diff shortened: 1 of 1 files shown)


JustOff:
I'd recommend you to study [code contribution guidelines](https://github.com/MoonchildProductions/UXP/blob/master/docs/CONTRIBUTING.md), in particular regarding [submitting pull requests](https://github.com/MoonchildProductions/UXP/blob/master/docs/CONTRIBUTING.md#general-guideline-for-submitting-pull-requests), and bring all of your current PRs in line with these rules.

wolfbeast:
Is it confirmed that this addresses #110 ? If not, then please don't use the `fixes` or `resolves` keywords with issue numbers because GitHub will auto-close them in that case.

alaviss:
I've updated the PR description. Hopefully it conforms to the guidelines.

Issue #361, opened 2018-05-16
Update Reader and Narrator components in Toolkit


Ascrod:
The Reader and Narrator components can be updated with changes from later FF versions. Some of the changes will include:
- Updating the Readability library
- Changing some functionality to use events instead of multi-process messaging
- Removing some telemetry and UI tour code
- Improving add-on access to reader and narrator APIs

Most of the work has actually already been done on my local copy, but I wanted to create an issue for this change before submitting a PR, as per the contribution guidelines.

mattatobin:
We can consider this a meta issue.

wolfbeast:
@Ascrod Since you updated these components, can you please check if [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1456652) has become applicable through your updates, and if so, patch it?

Ascrod:
The bug is applicable, and I believe the patch is too.  However, while using the [test site from the bug report](https://test.shhnjk.com/read_bypass.html), I found an issue with handling empty fragments in an href from within Reader View; this is preventing me from properly going through the repro steps for the first bug. I will look into both of these items.

Ascrod:
I have implemented a patch for handling local links properly (because it was not just an issue with empty fragments like I thought), followed by the patch from the bug report, but I'm seeing no difference in how cookies are being sent to the test site.

I did some testing with Firefox 61, using the repro steps outlined in the bug report. When the CLICK ME link is clicked in Reader View, a new tab opens to the cookie-checking page, still in reader view, and no cookies are sent. Upon exiting Reader View, only the "lax" cookie is sent. This, I assume, is the expected behavior.

While testing the same thing with Basilisk, when the new tab opens, before even exiting Reader View, both strict and lax cookies are sent in a request to the site. The cookies are sent again upon exiting Reader View. I do not think this is expected behavior.

Ascrod:
I am working on updating the Reader components again, and am also porting a few applicable patches over, including the following:
- [Bug 1399616](https://bugzilla.mozilla.org/show_bug.cgi?id=1399616)
- [Bug 1422680](https://bugzilla.mozilla.org/show_bug.cgi?id=1422680)
- [Bug 1151735](https://bugzilla.mozilla.org/show_bug.cgi?id=1151735)
- [Bug 1429442](https://bugzilla.mozilla.org/show_bug.cgi?id=1429442) - Involves some string additions that @JustOff will probably want to localize.

[...]

PR #367, opened 2018-05-16
Reader and Narrator Updates

Ascrod:
This updates the reader and narrator components in toolkit and both browser applications. See #361. Some of the more superficial changes from Mozilla were included with the intent of staying as close to the source as possible and making it easier to maintain going forward.


wolfbeast:
@Ascrod Since you updated these components, can you please check if [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1456652) has become applicable through your updates, and if so, patch it?

PR #369, opened 2018-05-17
external/nss: Don't link with libcrmf on --with-system-nss builds

alaviss:
Follow-up: #364 

This commit completely removes the `-lcrmf` for `--with-system-nss` builds, as libcrmf is not available on all systems (eg. Arch Linux).

This was meant to be a part of #364, but I forgot to push the commit to that PR. Sorry for the inconvenience.


mattatobin:
As a reminder, system libs are not supported for officially branded builds of our software and indeed could and will cause issues.

PR #395, opened 2018-05-27
Always build with SSE2 support & display an error if not available

trav90:
This PR enforces the use of SSE2 when building UXP with any version of GCC. In addition, if a user attempts to start either Pale Moon or Basilisk on a system that does not support SSE2, an error message should be displayed instead of the browser just silently failing to start.

I've tested and can confirm that SSE2 is being enforced on GCC builds on Linux, however I do not have any non-SSE2 hardware to test that the error message is displayed.


mattatobin:
I don't think this will override optimization flags in mozconfig... Does it?

trav90:
Yes it should override mozconfig flags because these flags are passed to the compiler after mozconfig flags, thus overriding them.

@wolfbeast what do you think? Enforce for all UXP applications or only for official branding? I’m fine with either approach; just want to prevent a slew of builds showing up that supports ancient hardware.

khronosschoty:
I no I was not asked, but in my opinion, it should at most only effect
officially branded builds.

wolfbeast:
I think this should apply regardless of build. UXP is a modern platform and should not be used to target very old hardware. There are more areas where this is or will be a problem as things are developed. It should not be left "up in the air" to break later on (which it will) by allowing it now.

SSE2 is and should be a hard requirement for all UXP applications.

trav90:
@wolfbeast in that case this is ready to merge.

mattatobin:
Doesn't this mean sse2 will override my specification of avx or avx2 when building and not cover this case at all?

trav90:
The sse2 flags should only override mozconfig flags that are contradictory (such as -mno-sse2); avx and other flags should not be overwritten.

I can't speak about ARM as I'm not familiar enough with it. Would have to do some research.

I can prepare a patch that will allow this to be overridden in mozconfig or we can use a Directive 4 solution. I'm fine with either.

mattatobin:
Your solution would be fine if the only applications were Basilisk and Pale Moon. They are not, however. Also, Directive 4 is only in effect for specific applications and only then for specific branding, namely ours.

It is also complemented by a private build configure flag to override that with transformative results like the text in the about boxes indicating its private build status.

My post above was directed to Moonchild more than you.

The bottom line is just like there must be balance between Developer and User choices with UXP there must be balance between Moonchild Productions and the rest of the world who would use the platform. Be it minor rebuilds, forks, or independent applications.

That is the reason, or so I thought, that UXP is a thing.

khronosschoty:
I agree.

wolfbeast:
I'm not sure why there is this much discussion. directive 4, Moonchild Productions applications vs. others etc. should not matter.
The solution must be one that applies regardless of application (so in `confvars.sh` is actually not a good place) that enforces SSE2 for the platform as a whole. A combination of filtering out unwanted disable flags and adding desired enable flags, applied for all platforms, would be best. This should be in the platform's root configuration somewhere, probably.

mattatobin:
I don't want to be limited to Pale Moon or Basilisk configuration. This means be it Borealis or XUL Runner or anything else I am forced to be SSE2. I shouldn't have to fork the platform because you are making product and marketing decisions for everyone.

trav90:
What about if we add a configure option (--enable-modern-simd or something similar) that could be enabled by default and would enforce the SSE2 flags be passed to the compiler. We could also tie in the warning message flag to this option, and use directive 4 to ensure that this option is always used when building with official branding.

It would allow people to override it in mozconfig, but maybe it would be an acceptable compromise?

mattatobin:
How it should work is if you want to assume SSE2 on an application basis then do so. If you want to be kind then ifdef it based on that..

Let Directive 4 enforce the optimization for it on official branding.

Minor rebuilds, forks of the code, and private builds should be able to do what they want with a minimum of imposed conditions. UXP is NOT mozilla-central or palemoon-central or basilisk-central.

wolfbeast:
So, you're insisting on wanting UXP to be somehow catering to pre pentium-IV hardware.
You do realize that this goes beyond just compiler optimizations, right, and that the SSE2 assumption **will** be made for hot paths in our development? I want to enforce SSE2 on the platform as a whole because it makes NO SENSE to keep catering to and targeting IA32 in 2018, and it will actually prevent some planned code improvements.

mattatobin:
Not really I prefer to use superior intruction set to SSE2 but if you enforce SSE2 then that overrides my ability to use AVX or newer.

Also, just how much is actually dependent on any arbitrary instruction set? That hasn't even been stated or made clear. To me it looks like 3rd party libs using assembly.

Now at some point in the future some of those libs may not have mmx or sse instructions and at that time it can be considered and justified then.

THIS is about impossing a restriction and indeed locking in of a specific instruction set without cause or justification beyond it is not how branded Basilisk or Pale Moon builds should be configured.

Assuming a position of SSE2 only for the entire platform for the same reasons as Windows XP needs to be justified.

wolfbeast:
You clearly misunderstood
it's a MINIMUM.
Since this is getting out of hand here, let's discuss on the forum.
https://forum.palemoon.org/viewtopic.php?f=62&t=19269

khronosschoty:
fwiw I had the same misunderstanding.

mattatobin:
Let me be absolutely crystal clear here if I specify to build as AVX or AVX2 will this as implimented still comple as SSE2?

khronosschoty:
I agree -- I am in favor of a build failing due to being miss-configured and not in favor of it being silently overridden.

wolfbeast:
Right, well that needs to be fixed then. it's not supposed to override -arch:AVX etc. This is why I made a point of saying to FILTER OUT as-needed.
The default for msvc is sse2. so it needs to filter out -arch:SSE and -arch:IA32.
And no, I do not agree with build failure, because then you'd have to put in extra checks EVERYWHERE

wolfbeast:
I guess I'll have to do this myself, then.

wolfbeast:
Doesn't look like this PR is even touching windows compilation. so what's all the screaming about?

mattatobin:
No you don't.. It is very simple to do it in a similar manor as I did for Directive 4 but NOT directive 4. It can be done from configure very early on in the process.

It isn't a problem and it can be a simple list of forbidden optimization flags.

Would this be acceptable?

khronosschoty:
It would be very much appreciated if it was handled in the manner suggested above (by Tobin).

wolfbeast:
Go ahead Tobin.

mattatobin:
Incompatible optimizations is a configure error.

wolfbeast:
Both have been reverted.

trav90:
Was this partially re-landed? I noticed it in the commit history this morning.

JustOff:
I was also surprised when I saw that `mach` suddenly started a complete rebuild, but I thought that it was already discussed somewhere and I just missed it.

wolfbeast:
Yes I partially re-landed this for GCC (adding the flags for sse2 in the default compiler config). The other solution would not enforce anything, just prevent unwanted flags.

mattatobin:
So does that override flags in MOZ_OPTIMIZE?

wolfbeast:
Now, before getting in your ivory tower again you could have taken a NORMAL approach and simply ask. Or better yet, look at how GCC handles these command-line flags when specified, before accusing me of all sorts of things. MSVC might override, indeed, but GCC does not. `-msse2` and other `-m` flags like it only **enable** the use of the instruction sets (with `no` variants disabling them). You think I didn't look this up? 

I'm tired of this, Tobin. Every single time you think you have a solution it has to be exactly your solution and nothing else without a ton of infighting or needing exact control over anyone doing anything that is even within a mile of "your" code. We discussed this before, and you even said you were fine with these flags as a minimum so what the fuck are you doing getting in my hair when I make sure sse2 is used? And no, that is NOT branding dependent.

Your solution prevents people overriding the minimum with `no` optimization flags, which is perfect, but it doesn't actually set the use of the instruction set. Requiring that people manually add it to their optimize flags (what you apparently "intended" but never saw fit to let anyone know -- I'm still not clairvoyant) is backwards.

wolfbeast:
Also
> No, because no one right now knows the behavior when both are specified..

Read the gcc manual. It's all in there. https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html#x86-Options

mattatobin:
So what does "depresses" mean that the SSE2 instructions are still there but it will add and use AVX for AVX2 instead? Or does it mean it makes -msse2 like it isn't present? "When needed" when is that?

When I specify a valid -m flag I expect it to be the only one in play.. So is this the case?

khronosschoty:
> Yes I partially re-landed this for GCC (adding the flags for sse2 in the default compiler config). The other solution would not enforce anything, just prevent unwanted flags.

That would be a better solution. Anyway I'm going to stay out of this now.

mattatobin:
Know what.. That's fine.. I'm done.. Do what you are gonna do then.. UXP belongs to you as your huge declairation on the readme says... I don't care anymore.

The platform aka moonchild-central isn't my consern anymore. I have more pressing things to do than micromanage this codebase I slaved over for four months.

I am now hands off in every respect outside of Phoebus and when that is done so am I.

wolfbeast:
> I am now hands off in every respect outside of Phoebus and when that is done so am I.
Okay. Your choice and your freedom. I appreciate you're willing to stick around to bring Phoebus to completion instead of ragequitting. Thank you.

wolfbeast:
> So what does "depresses" mean? That the SSE2 instructions are still there but it will add and use AVX for AVX2 instead? Or does it mean it makes -msse2 like it isn't present? "When needed" when is that?
> 
> When I specify a valid -m flag for instriction set most expect it to be the only one in play.. So is this the case?

To answer these questions: AVX is a special case because AVX is mutually exclusive with SSE*. All these flags in GCC only flag to the compiler "You are allowed to use these instructions", and in practice this means the "highest" of the class is generally used when applicable to specific code. _When_ they are used is determined _by the compiler_ based on the code being compiled (ask GCC devs for details). In the case of AVX, the "depresses" statement, if I read correctly, acts like `-msse*` isn't specified, in effect overriding them and always using AVX instructions when both classes of instructions are possible. All the other flags are simply "You can use this if you want to" for specific pieces of compiled code.

PR #397, opened 2018-05-27
Only display SSE2 error dialog on 32-bit Linux

trav90:
Just what it says in the tin. All 64-bit CPU's support SSE2 so there's no reason to have this code on 64-bit builds.

Tested on both 32 and 64-bit environments and works as intended.


trav90:
Follow up to #395

Issue #65, opened 2018-03-14
Use the Preprocessor instead of AppConstants


mattatobin:
*CLONE OF https://github.com/MoonchildProductions/moebius/issues/156*

Mozilla has taken it upon themselves to make what are build-time configuration options into run-time evaluation through the use of AppConstants.

This is problematic:
- It inflates target code size
- It increases complexity with more than a 1000 individual run-time checks
- It reduces performance because of said checks
- It makes no difference for parts where run-time checks can't be used and preprocessing is needed anyway, creating a big mess of code that is half build-time, half run-time.

We need to return to preprocessing and reducing object code size that way. Preprocessors exist for a good reason.
If we want UXP to be usable for Pale Moon and other applications, this is likely going to be a hard and critical requirement.

Part of this work will be moving configure defines back to makefile configure from the pythonification `moz.configure`. 

To start off, we'll need to revert the AppConstants in `toolkit\modules\AppConstants.jsm` and the sites where these are used to normal preprocessor `#ifdef`s; probably one at a time to be safe and not cause hard-to-trace breakage.

mattatobin:
I am not going to modify tests or the following directories:

- `devtools/`
- `toolkit/telemetry`

mattatobin:
That should be all of them except as noted above here in UXP.

Just waiting on @wolfbeast to give the latest ones a cursory lookover.

mattatobin:
Taking this out of the Milestone because `Basilisk` and `BinOC Mail/Calendar` can be done whenever.

mattatobin:
Current state of this issue: DevTools and MozTests can piss off, Telemetry can die, and all there is left is the `Basilisk` and `Interlink` which can be done at any time.. So I am gonna close this issue.

**RESOLVED GOODENOUGH**

PR #422, opened 2018-05-31
Get rid of AppConstants in `application/palemoon`

JustOff:
`AppConstants.jsm` is used in the following two places in `application/palemoon`:

* shell component
* autocomplete.xml

This PR uses preprocessing to get rid of it.

Tag #65.

### DIFF
--- a/application/palemoon/components/shell/ShellService.jsm
+++ b/application/palemoon/components/shell/ShellService.jsm
@@ -8,7 +8,6 @@ this.EXPORTED_SYMBOLS = ["ShellService"];
 
 const { classes: Cc, interfaces: Ci, utils: Cu, results: Cr } = Components;
 
-Cu.import("resource://gre/modules/AppConstants.jsm");
 Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "WindowsRegistry",
@@ -26,20 +25,18 @@ var ShellServiceInternal = {
    * environments.
    */
   get canSetDesktopBackground() {
-    if (AppConstants.platform == "win" ||
-        AppConstants.platform == "macosx") {
-      return true;
+#if defined(XP_WIN) || defined(XP_MACOSX)
+    return true;
+#elif XP_LINUX
+    if (this.shellService) {
+      let linuxShellService = this.shellService
+                                  .QueryInterface(Ci.nsIGNOMEShellService);
+      return linuxShellService.canSetDesktopBackground;
     }
-
-    if (AppConstants.platform == "linux") {
-      if (this.shellService) {
-        let linuxShellService = this.shellService
-                                    .QueryInterface(Ci.nsIGNOMEShellService);
-        return linuxShellService.canSetDesktopBackground;
-      }
-    }
-
     return false;
+#else
+    return false;
+#endif
   },
 
   /**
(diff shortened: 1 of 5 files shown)


application/palemoon/components/shell/ShellService.jsm
```diff
@@ -26,20 +25,18 @@ var ShellServiceInternal = {
...
-
-    if (AppConstants.platform == "linux") {
-      if (this.shellService) {
-        let linuxShellService = this.shellService
-                                    .QueryInterface(Ci.nsIGNOMEShellService);
-        return linuxShellService.canSetDesktopBackground;
-      }
+#if defined(XP_WIN) || defined(XP_MACOSX)
+    return true;
+#elif XP_LINUX
```
adisib: Shouldn't this be "#elif defined(XP_LINUX)" ?

mattatobin: No.. one does not have to do that for one codition.

adisib: I meant in terms of consistency with how it is done elsewhere in the codebase. But I guess that is just a personal nitpick and not worth bringing up.

JustOff:
Please put this on hold for a while, I have to re-check something.

JustOff:
Could someone tell me why Mozilla uses such a complex method to detect `android` and `linux` platforms?

https://github.com/MoonchildProductions/UXP/blob/ac25827a87d86f1cf9e48aab6605f77a2c89041a/toolkit/modules/AppConstants.jsm#L147-L162

Why the detection uses `MOZ_WIDGET_GTK`, `MOZ_WIDGET_ANDROID` and `XP_LINUX`, instead of just check
```
#elif ANDROID
  "android",
#elif XP_LINUX
  "linux",
```
Do I also need to build a complex rule or `#if defined(XP_LINUX) && !defined(ANDROID)` is enough?

wolfbeast:
Android and MacOSX are both Linux variants that have `XP_LINUX` set. the `ANDROID` define is only set in nss, and the most definitive way to check for specific OS is to check the widget decision.
Look at the conditionals - it needs that order. If you have GTK widgets, it's always a variant of pure Linux. if XP_MACOSX is set, you're on the Mac flavor, if you have Android widgets, you're on Android. If you pass all those checks and XP_LINUX is still set, then it's a non-GTK flavor of pure Linux. You can't simplify this without breaking things.

JustOff:
Well, thank you for the clarification. But in this case, is not it better to continue using AppConstants, at least for platform detection?

JustOff:
One more question, is it possible that `MOZ_WIDGET_GTK` is defined, while `XP_LINUX` is not?

JustOff:
> You're not here to optimize appConstants, you're here to do away with them ;)

Sorry, but I can't work with the code not trying to understand and optimize it. And in this case the problem is aggravated by the fact that if you literally follow the `AppConstants` rules for detecting the platform, you will get an ugly code that does not occur anywhere else in the sources.

> Android and MacOSX are both Linux variants that have XP_LINUX set.

I think you were mistaken, they are variants of `XP_UNIX`, but not `XP_LINUX`, see [init.configure#L527-L570](https://github.com/MoonchildProductions/UXP/blob/36f73c8cd27e62cbd3e85939d6fe11a240e3416f/build/moz.configure/init.configure#L527-L570).

So, the question of whether it is possible that `MOZ_WIDGET_GTK` is defined, while `XP_LINUX` is not remains very relevant in order to move on. Perhaps @mattatobin or @trav90 can add something about this?

wolfbeast:
It does mean, however, that if you see an appconstants check with "linux" in it, that you need to check what it was before in Mozilla code, because it could be either of `MOZ_WIDGET_GTK` or `XP_LINUX` that you have to change it to.

wolfbeast:
Also, why are you even looking at appconstants.jsm if all you are doing here is removing the use of them from a specific part of the tree? Unless you intend to remove specific appconstant values from the entire tree, you should leave that module alone.

mattatobin:
Here is a simple resolution. No one touches AppConstants but me. Stick to what ever form is used in the file at hand in the codebase and I will work through converting it back to preprocessing over time.

wolfbeast:
I'll close this PR then

JustOff:
Ok, deal.

Issue #239, opened 2018-04-23
Remove UITour


wolfbeast:
UITour has a bit of integration with Reader View so care must be taken. The rest seems fairly straightforward.

PR #424, opened 2018-05-31
Remove UITour

Ascrod:
Addresses #239. Also removes customization tips and tracking protection tour.


browser/modules/SelfSupportBackend.jsm
```diff
@@ -90,8 +86,7 @@ var SelfSupportBackendInternal = {
     }
 
     // Make sure UITour is enabled.
-    let uiTourEnabled = Preferences.get(PREF_UITOUR_ENABLED, false);
-    if (!uiTourEnabled) {
+    if (true) {
```
wolfbeast: Never do this.
If you have an if statement that is always true, remove the if statement.

Ascrod:
Removed the condition, sorry about that. Note that this will generate a warning about unreachable code.

Since it appears that this feature requires UITour to work, we should probably look into what it does and whether it needs to be modified or removed.

wolfbeast:
Then you also need to remove the unreachable code because it would only be in use in case UITour exists.

JustOff:
And why do we need `SelfSupportBackend.jsm` at all? Is there anything useful left after you removed addObserver call?

Ascrod:
I think you're right. The module is only referenced in nsBrowserGlue.js by calls to the (now-broken) init and the uninit methods. Should I remove it with this PR?

wolfbeast:
> At first glance it seems that we can get rid of it entirely.

That would be for a follow-up bug. Please keep the contributing guidelines in mind. Things may follow logically one after the other, but we need to keep one github issue for one thing.

wolfbeast:
> (now-broken) init

What's broken? does it need fixing before merge?

wolfbeast:
Dead code in our tree is nothing new. Cleaning that up, as said, should be its own thing.

PR #429, opened 2018-06-01
Check if there are any unpreprocessed files in '@MOZ_OBJDIR@/dist/bin'

### LINKED ISSUE #428: Use a final dist check to catch any unpreprocessed files.
Because some want to have this discussion again.

JustOff:
This is intended to reduce the number of errors in case not all the necessary files are added to the preprocessing.

The check is performed automatically after each `mach build`.

Resolves #428.

### DIFF
--- a/python/mozbuild/mozbuild/mach_commands.py
+++ b/python/mozbuild/mozbuild/mach_commands.py
@@ -541,9 +541,11 @@ class Build(MachCommandBase):
                 # as when doing OSX Universal builds)
                 pass
 
-        # Check if there are any unpreprocessed files
-        grepcmd = 'grep -E -r "^(#|%)ifdef" --include=\*.{js\*,css,x\*,h\*,manifest,dtd,properties} '\
-                  + self.topobjdir + '/dist/bin'
+        # Check if there are any unpreprocessed files in '@MOZ_OBJDIR@/dist/bin'
+        # See python/mozbuild/mozbuild/preprocessor.py#L293-L309 for the list of directives
+        grepcmd = 'grep -E -r "^(#|%)(define|el|endif|error|expand|filter|if|include|literal|undef|unfilter)" '\
+                  + '--include=\*.{css,dtd,h\*,js\*,x\*,manifest,properties,rdf} '\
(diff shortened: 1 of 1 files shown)


python/mozbuild/mozbuild/mach_commands.py
```diff
@@ -541,9 +541,11 @@ def build(self, what=None, disable_extra_make_dependencies=None, jobs=0,
                 # as when doing OSX Universal builds)
                 pass
 
-        # Check if there are any unpreprocessed files
-        grepcmd = 'grep -E -r "^(#|%)ifdef" --include=\*.{js\*,css,x\*,h\*,manifest,dtd,properties} '\
-                  + self.topobjdir + '/dist/bin'
+        # Check if there are any unpreprocessed files in '@MOZ_OBJDIR@/dist/bin'
+        # See python/mozbuild/mozbuild/preprocessor.py#L293-L309 for the list of directives
+        grepcmd = 'grep -E -r "^(#|%)(define|el|endif|error|expand|filter|if|include|literal|undef|unfilter)" '\
```
wolfbeast: What is `#el` ?

JustOff: `#el` catches both `#else` and `#elif`, just like `#if` does with `#(if|ifdef|ifndef)`.

wolfbeast: OK, seems I have to repeat myself again today.

Only check for `#endif` to catch any `if` constructs. We're not trying to sanity check the preprocessor directives here, just checking if they are still there. If not processed, any `if` construct will ALWAYS have an `#endif` somewhere. Checking for `#endif` is enough, any other directive part of `if` constructs don't have to be checked.

JustOff: Of course I saw what you wrote, but I didn't think it was a requirement, but just a suggestion. Since you insist, I'll update the regex by removing `if` and `el`, however I honestly don't see any difference in terms of result or performance impact.

wolfbeast: I insist because short match strings are likely to trigger false positives. It's not just about performance.

PR #460, opened 2018-06-06
backport mozbug #546387: fix assert failure in gtk which prints to the console

g4jc:
GNU/Linux users see "Gtk-CRITICAL **: gtk_clipboard_set_with_data: assertion 'targets != NULL' failed" spammed to the console when selecting null data on the newtab page and other areas. This fixes that issue by backporting [mozbug 546387](https://hg.mozilla.org/releases/mozilla-esr52/raw-rev/d13e3fefb76e).


wolfbeast:
You have to wonder why it took 7 years for a 1-liner fix to land...

Issue #447, opened 2018-06-05
Keep a running update of HSTS preload lists.


wolfbeast:
UXP should keep updated with somewhat recent HSTS preload lists.

@trav90  has offered to do this.

(responsibility tracking issue, leave open)

trav90:
I am going to look into automating this on a dedicated machine to keep this list updated a bit more regularly.

wolfbeast:
Since HSTS is a long-term commitment thing, once every fortnight should be frequent enough -- I wouldn't want to have near-daily updates like Mozilla does, that's just a waste and commit noise we don't want.

wolfbeast:
@trav90 obviously doesn't have time to do this anymore and I don't think of running these updates with any sort of regularity either. We either need someone willing to run this regularly and committing these updates to the tree, or simply do away with the preload list altogether which is becoming less effective as it is anyway.

wolfbeast:
Closing this because #1498 means we are no longer in need of updating.

PR #461, opened 2018-06-07
Improve HSTS preload list generation

### LINKED ISSUE #454: Improve HSTS preload list generation
We've made some improvements to the HSTS preload generation list in Tycho that should also be ported here. In addition, we should revert a Mozilla change (made primarily because of e10s) and return to the readable, clearly-delimited preload list that Tycho uses instead of Mozilla's blob of characters.

I'll be taking this and regenerating the preload list from scratch once done.

- [x] Drop a site from the preload list if there is an error connecting to the host (port from Tycho).
- [x] Increase number of concurrent lookups in generation script to 15 (port from Tycho).
- [x] Restore the clearly-delimited format of the HSTS list (as is currently used in Tycho).
- [x] Regenerate HSTS preload list from scratch.

Blocks #447

trav90:
Resolves #454. Also tag #447.

This PR ports the HSTS preload generation script improvements from Tycho and regenerates the HSTS preload list from scratch.


trav90:
I've also added a commit that updates the HSTS pref name used by Pale Moon to match that of the platform. This fixes the HSTS box in Preferences so that it actually works and enabled/disables HSTS per user choice (enabled by deafult).

PR #463, opened 2018-06-08
[PALEMOON] Add missed strings required by page info

JustOff:
This fixes following error in `pageInfo.xul`:
```
NS_ERROR_FAILURE: Component returned failure code: 0x80004005 (NS_ERROR_FAILURE) [nsIStringBundle.GetStringFromName]  XStringBundle:22
	getString XStringBundle:22:22
	securityOnLoad chrome://browser/content/pageinfo/security.js:282:12
	makeGeneralTab chrome://browser/content/pageinfo/pageInfo.js:550:3
	loadPageInfo chrome://browser/content/pageinfo/pageInfo.js:356:3
	loadTab chrome://browser/content/pageinfo/pageInfo.js:441:3
	onLoadPageInfo chrome://browser/content/pageinfo/pageInfo.js:337:3
	onload chrome://browser/content/pageinfo/pageInfo.xul:1:1
```

Tag #121.

### DIFF
--- a/security/manager/locales/en-US/chrome/pippki/pippki.properties
+++ b/security/manager/locales/en-US/chrome/pippki/pippki.properties
@@ -98,7 +98,8 @@ clientAuthStoredOn=Stored on: %1$S
 # Page Info
 pageInfo_NoEncryption=Connection Not Encrypted
 pageInfo_Privacy_None1=The website %S does not support encryption for the page you are viewing.
-pageInfo_Privacy_None2=Information sent over the Internet without encryption can be seen by other people while it is in transit. 
+pageInfo_Privacy_None2=Information sent over the Internet without encryption can be seen by other people while it is in transit.
+pageInfo_Privacy_None3=The page you are viewing is not encrypted.
 pageInfo_Privacy_None4=The page you are viewing was not encrypted before being transmitted over the Internet.
 # LOCALIZATION NOTE (pageInfo_EncryptionWithBitsAndProtocol and pageInfo_BrokenEncryption):
 # %1$S is the name of the encryption standard,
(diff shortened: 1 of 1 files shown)


security/manager/locales/en-US/chrome/pippki/pippki.properties
```diff
@@ -98,7 +98,8 @@ clientAuthStoredOn=Stored on: %1$S
 # Page Info
 pageInfo_NoEncryption=Connection Not Encrypted
 pageInfo_Privacy_None1=The website %S does not support encryption for the page you are viewing.
-pageInfo_Privacy_None2=Information sent over the Internet without encryption can be seen by other people while it is in transit. 
+pageInfo_Privacy_None2=Information sent over the Internet without encryption can be seen by other people while it is in transit.
+pageInfo_Privacy_None3=The page you are viewing is not encrypted.
```
mattatobin: I think Pale Moon should switch from pageInfo_Privacy_None3 to pageInfo_Privacy_None4 because it is a superior string

security/manager/locales/en-US/chrome/pippki/pippki.properties
```diff
@@ -110,6 +111,7 @@ pageInfo_Privacy_Encrypted1=The page you are viewing was encrypted before being
 pageInfo_Privacy_Encrypted2=Encryption makes it difficult for unauthorized people to view information traveling between computers. It is therefore unlikely that anyone read this page as it traveled across the network.
 pageInfo_MixedContent=Connection Partially Encrypted
 pageInfo_MixedContent2=Parts of the page you are viewing were not encrypted before being transmitted over the Internet.
+pageInfo_Privacy_Broken1=Parts of the page you are viewing were not encrypted or the encryption is not strong enough before being transmitted over the Internet.
```
mattatobin: Pale Moon should switch to pageInfo_MixedContent2 but all Applications should use the string contents of pageInfo_Privacy_Broken1 AS pageInfo_MixedContent2 in my view..

@wolfbeast ?

JustOff: Please pay attention, that Basilisk and Pale Moon have different logic in `pageinfo/security.js`, and only Basilisk distinguishes weak ciphers in mixed content. So I'm not sure we can use the common string here.

security/manager/locales/en-US/chrome/pippki/pippki.properties
JustOff: I'm not sure the wordings are equivalent, but if everyone agrees, I also won't mind.

security/manager/locales/en-US/chrome/pippki/pippki.properties
mattatobin: Maybe Pale Moon should too :P

JustOff: I'd prefer first to deal with missed strings and consider improving of page info as a separate task.

mattatobin: Sounds reasonable.

PR #464, opened 2018-06-09
[suggestion] Fix build bustage in services\sync\modules\service.js

janekptacijarabaci:
Throws an build error:
```
mozbuild.preprocessor.Error:
('\\services\\sync\\modules\\service.js', 1070, 'SYNTAX_ERR',
"0 // FSyncMS doesn't support this and it will break auth if left enabled!")
```


janekptacijarabaci:
Implemented:
https://github.com/MoonchildProductions/UXP/commit/16e2a93396f244ae39b91e039ba846eebf3d4237

PR #467, opened 2018-06-09
Hide tracking protection controls from preferences UI

### LINKED ISSUE #425: Tracking protection blocklists not displayed in Basilisk preferences
When Basilisk is built with safe browsing disabled, several constants in [blocklists.js](https://github.com/MoonchildProductions/UXP/blob/271c7b598d2daa42a8cc25ca72ecb538953bcfd8/browser/components/preferences/blocklists.js#L9) are ifdef'd out but are still used in the code. Thus, when the user goes to change what blocklist they're using for tracking protection (Preferences > Privacy > Tracking > Change Block List), the list of available block lists fails to load.

trav90:
This commit hides the tracking protection controls from the UI (while leaving the Do Not Track option there). I also removed the mention of tracking protection from the Do Not Track comment in preferences as Do Not Track still works even without tracking protection enabled.

Resolves #425.

### DIFF
--- a/application/basilisk/components/preferences/in-content/privacy.xul
+++ b/application/basilisk/components/preferences/in-content/privacy.xul
@@ -117,8 +117,7 @@
       </vbox>
     </hbox>
   </vbox>
-  <vbox id="trackingprotectionpbmbox">
-    <caption><label>&tracking.label;</label></caption>
+  <vbox id="trackingprotectionpbmbox" hidden="true">
     <hbox align="center">
       <checkbox id="trackingProtectionPBM"
                 preference="privacy.trackingprotection.pbmode.enabled"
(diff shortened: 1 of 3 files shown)


JustOff:
Perhaps I missed something, but why are you going to hide the controls and change the strings regardless of whether we build safebrowsing or not?

wolfbeast:
Just yanking them out? I think we should just make this conditional on `MOZ_SAFEBROWSING` or so.

trav90:
To be honest I don't understand enough of how the preprocessor works to be able to add the ifdef conditionals. I spent a couple hours on this but only managed to completely break the preferences page when adding the conditionals.

trav90:
@JustOff I’m sure it is something simple but you’re welcome to take it if you’d like.

Issue #230, opened 2018-04-22
Make safebrowsing optional at build time


wolfbeast:
The reliance on Google Safebrowsing is no longer configurable at build time and currently purely controlled with preferences (i.e.: always built/included).

This needs to be made a build-time decision again.

trav90:
I'll take this one.

trav90:
Gonna need another set of eyes on this one. I've restored the configure option and most of the #ifdef's for safebrowsing (still need to make another pass through and make sure I didn't miss any areas that still need them), but in the process I've completely broken the "in-content" preferences page (about:preferences) in Basilisk. The error console throws a warning about an invalid character [here](https://github.com/MoonchildProductions/UXP/commit/41d68e3f70b0aabc01cf9a49c43cf1aab0f4bdce#diff-0cc486f5fbab6f1b1f8b2ee1fd60adb6R26).

What is confusing me is that other in-content .js files are using #ifdef's and no errors are being throws (such as the one here: http://xref.palemoon.org/uxp-trunk/source/browser/components/preferences/in-content/content.js#81). So I'm not sure what it is that I'm overlooking.

The work I've done thus far is here: https://github.com/trav90/UXP/tree/safebrowsing-work

JustOff:
It seems you simply missed to include `security.js` into pre-processing [here](https://github.com/MoonchildProductions/UXP/blob/41d68e3f70b0aabc01cf9a49c43cf1aab0f4bdce/browser/components/preferences/in-content/jar.mn#L17).

trav90:
Thanks @JustOff I knew it had to be something simple I was overlooking. I'll include that file in pre-processing and give it another go.

trav90:
Thanks @JustOff that did the trick. This should now be mostly done. I’ll submit a PR once I’ve had a chance to go back over it and properly test.

PR #469, opened 2018-06-10
[BASILISK] Remove tracking protection controls from the UI if safebrowsing isn't built

### LINKED ISSUE #425: Tracking protection blocklists not displayed in Basilisk preferences

JustOff:
If safebrowsing isn't built:

* By default disable tracking protection in Private Browsing mode 
* [BASILISK] Remove tracking protection controls from the UI
* [BASILISK] Remove tracking protection title, controls and description from aboutPrivateBrowsing

Tag #230, resolves #425.

### DIFF
--- a/application/basilisk/components/preferences/in-content/jar.mn
+++ b/application/basilisk/components/preferences/in-content/jar.mn
@@ -8,7 +8,7 @@ browser.jar:
    content/browser/preferences/in-content/subdialogs.js
 
    content/browser/preferences/in-content/main.js
-   content/browser/preferences/in-content/privacy.js
+*  content/browser/preferences/in-content/privacy.js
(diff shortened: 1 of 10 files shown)


wolfbeast:
Let me know when you think this is ready to be merged.

JustOff:
I think it is. Sorry for the multiple updates to the open PR.

PR #482, opened 2018-06-11
Improve performance of cross-compartment wrapper cleanup.

### LINKED ISSUE #208: Improvements to cross-compartment wrapper handling when managing tabs
Mozilla has resolved this bug with different final implementation and Tycho/Pale Moon currently uses initial implementation.

Would it be possible to backport Mozilla final implementation (Mozilla 55) to UXP?

Bug 816784 - Lots of time spent in WindowDestroyedEvent::Run/js::NukeCrossCompartmentWrappers when closing tabs/windows

https://bugzilla.mozilla.org/show_bug.cgi?id=816784

wolfbeast:
This is the batching approach as used in Tycho, refactored to UXP.

Could use a run and/or review of code. Been staring at this puzzle piece for 5 hours trying to make it fit.
This resolves #208.

### DIFF
--- a/dom/base/nsGlobalWindow.cpp
+++ b/dom/base/nsGlobalWindow.cpp
@@ -67,6 +67,7 @@
 #include "mozilla/Likely.h"
 #include "mozilla/Sprintf.h"
 #include "mozilla/Unused.h"
+#include "../../js/public/HashTable.h"
 
 // Other Classes
 #include "mozilla/dom/BarProps.h"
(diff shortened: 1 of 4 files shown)


JustOff:
Unfortunately I'm not so familiar with this part of the platform code, and also not quite understand what I can check by running the browser with these changes.

wolfbeast:
Open a shit ton of tabs, then exit the browser. See how long it takes for the browser to completely shut down. With this change in, it should be considerably shorter.

JustOff:
I'm building with this PR applied to the master and going to find any difference.

JustOff:
Well, I'd say I can't see any difference in the browser shutdown time. I my case a bunch of 49 tabs of misc shit closes in about 9 seconds with and without this PR applied.

wolfbeast:
With 50 tabs you won't see much difference by just winging it. This is for the situation where people have several hundred tabs open, often with many overlapping pages of the same site.

SpockMan02:
Do the tabs need to be closed, or is quitting/exiting enough even if the session is kept on next startup?

wolfbeast:
This is about the cleanup of cross-compartment wrappers when shutting down or closing a window with many tabs. You don't have to close the individual tabs (in fact that would not trigger the code) but a shutdown of the browser (keeping the session) should see improvement with a huge amount of tabs present.

SpockMan02:
Quitting with a little over 500 tabs takes about four seconds, both with Pale Moon built from 660042b0a93eccc59980fbfb6c89787018b652ba and after pulling `CCW-perf`. Tested on macOS 10.12.

wolfbeast:
Okay, so it's not even an issue, then. I'll hold off on this then because it adds unnecessary complexity.

Issue #483, opened 2018-06-11
Memory leak in the WebP decoder


SorceressEpele:
After closing a memory-demanding tab (Such as a page on giphy after scrolling down considerable length), the browser keeps a persistently high memory use (In this case, I'm sitting at 3.4GB even after using the built-in about:memory to try and reclaim some.)

Tested in Version: 28.0.0a4 (64-bit) (2018-06-10)
Reproduction: Scroll down on giphy.com for a bit.
Expectation: More memory from that tab to be reclaimed after closing.
Reality: Claimed back just 1.5GB of memory (4.9GB Peak down to 3.4GB after closing tab)

wolfbeast:
Since you're familiar with about:memory, it would be nice to know where the memory report says the memory is allocated (which are the large consumers?)

JustOff:
I confirm this, in my case the memory not used anymore after closing the 2GB tab with https://giphy.com stuck in the heap-unclassified:

```
Pale Moon:

1,432.45 MB (100.0%) -- explicit
├──1,351.01 MB (94.31%) ── heap-unclassified

Basilisk:

1,291.93 MB (100.0%) -- explicit
├──1,195.19 MB (92.51%) ── heap-unclassified
├─────38.17 MB (02.95%) -- heap-overhead
```

At the same time, I can't even get Firefox ESR 52 with e10s disabled to take more than 512 MB by scrolling https://giphy.com, and all memory is returned when I close the tab.

JustOff:
Here is the basic bisection results:

Browser | Date | Result
------------------ | --- | -- 
Basilisk UXP [2018.06.01](https://github.com/MoonchildProductions/UXP/releases/tag/v2018.06.01) | 2018.06.01 | Problem
Basilisk UXP [CP1](https://github.com/MoonchildProductions/UXP/tree/FullFunction_CP1) | 2018.02.11 | Problem
... | ... | ?
Basilisk UXP [2018.02.02](https://github.com/MoonchildProductions/UXP/tree/c28c5b704fb3f3af6e7846abd73f63da1e35921f) | 2018.02.02 | Ok
Firefox ESR 52.6.0 | 2018.01.19 | Ok
Basilisk Moebius | 2017.12.01 | Problem

JustOff:
Well, it seems I found the culprit: it's a WebP image decoder, and with `image.webp.enabled -> false` the problem goes away.

wolfbeast:
[...]

PR #489, opened 2018-06-12
Disable WebP decoder until memory leak is resolved

JustOff:
Just what it says in the tin. Tag #483.


wolfbeast:
Considering the memory leak is resolved I'll back this out.

PR #500, opened 2018-06-15
[BASILISK] Add UA overrides for YouTube Live

### LINKED ISSUE #499: YouTube live chat box doesn't work
The chat box on YouTube live streams doesn't work due to UA discrimination.

trav90:
Resolves #499.

### DIFF
--- a/application/basilisk/branding/shared/uaoverrides.inc
+++ b/application/basilisk/branding/shared/uaoverrides.inc
@@ -24,6 +24,8 @@ pref("@GUAO_PREF@.accounts.firefox.com", "Mozilla/5.0 (@OS_SLICE@ rv:@GK_VERSION
 pref("@GUAO_PREF@.addons.mozilla.org", "Mozilla/5.0 (@OS_SLICE@ rv:@GK_VERSION@) @GK_SLICE@ @FX_SLICE@");
 
 // Required for domains that have proven unresponsive to requests from users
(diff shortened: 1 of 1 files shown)


application/basilisk/branding/shared/uaoverrides.inc
```diff
@@ -24,6 +24,8 @@ pref("@GUAO_PREF@.accounts.firefox.com", "Mozilla/5.0 (@OS_SLICE@ rv:@GK_VERSION
 pref("@GUAO_PREF@.addons.mozilla.org", "Mozilla/5.0 (@OS_SLICE@ rv:@GK_VERSION@) @GK_SLICE@ @FX_SLICE@");
 
 // Required for domains that have proven unresponsive to requests from users
+pref("@GUAO_PREF@.youtube.com","Mozilla/5.0 (@OS_SLICE@ rv:42.0) @GK_SLICE@ Firefox/42.0 @APP_SLICE@");
+pref("@GUAO_PREF@.gaming.youtube.com","Mozilla/5.0 (@OS_SLICE@ rv:42.0) @GK_SLICE@ Firefox/42.0");
```
wolfbeast: Are you sure using 42.0 is the right choice for UXP? Maybe we should use 60.0...

JustOff: I'd definitely prefer to use 42.0 that enforces old design as long as it's possible.

wolfbeast: Right; I forgot about that.
Good to go then!

PR #501, opened 2018-06-15
[PALEMOON] Remove UA override for Google Fonts

trav90:
This should no longer be necessary ([issue 749](https://github.com/MoonchildProductions/Pale-Moon/issues/749) on the PM repo is labelled "Solved by UXP").

While it shouldn't be needed anymore I have not tested this due to not currently having access to a Linux system with a GUI.

@khronosschoty @JustOff  would one of you be able to test this for me while I'm away?


JustOff:
What is the best way to verify this?

khronosschoty:
I will attempt to test this.

wolfbeast:
Ah yes unicode ranges are now fully supported on Linux so that can go. A quick test is appreciated :)

khronosschoty:
So I compiled back tracked through as best I could to see if I could reproduce the original issue but I can't. I tested the original issue reported in the forum on Pale Moon 27.0.1 and could not reproduce the issue there either. So in short I think my tests are of limited value.

wolfbeast:
Maybe Google actually shaped up (miracle if so, probably not because of us).
I'm just going to assume this is good without an explicit test.

Issue #317, opened 2018-05-01
[PALEMOON] Win8+ close icons incorrectly mapped


Lootyhoof:
On Windows 8 and above, as we use `toolkit` to style the close buttons (taken primarily from `Basilisk`), we also inherit the Win8+ close buttons that `Basilisk` uses. This both results in a different style of close buttons to what one would expect from Pale Moon, as well as inactive tab close buttons being mapped wrongly (due to expecting the old icons). All other icon mappings are as expected due to `.close-icon` being styled by `toolkit`.

![Win8+ close buttons](https://i.imgur.com/IyG1BVV.png)

It's possible to fix the mapping of the current icons in CSS (though we'd need to deal with the HiDPI `@2x` icons too), but do we actually want these icons on Pale Moon?

Lootyhoof:
Tag #121

wolfbeast:
Just grab them from tycho and map them locally in the application skin? Not sure why this was put in toolkit anyway since it's part of the main application ui.

Lootyhoof:
It's not just used for the tabs though, it's used for *any* close icons. This is why it's in toolkit.

We could place the close icons from Tycho and put them in the Pale Moon `browser` skin, sure, but since not all close icons live directly in the browser UI, this would create an inconsistency.

wolfbeast:
There's more that should, I just haven't gotten around to classifying the various bugs yet.

wolfbeast:
@Lootyhoof isn't it possible to override toolkit styling with application styling?

Lootyhoof:
I've been thinking about this. If we use `application={ec8030f7-c20a-464f-9b0e-13a3a9e97384}` in [jar.mn](https://github.com/MoonchildProductions/UXP/blob/master/toolkit/themes/windows/global/jar.mn), we could basically make Basilisk use its own `global.css` and we can just use a Pale Moon-specific one without the different Windows 8 icons (but basically otherwise the same as Basilisk's). This way, we could potentially add Pale Moon's older close icons (i.e. pre-28) and just rename the Basilisk ones and change the overrides as required (and add some more lines for the Win8 icons in the process). Pale Moon's icons and `global.css` could become the platform default in that case.

We could still do HiDPI if you have higher-res versions of those icons too. That would need work in `browser.css` too though to account for it.

wolfbeast:
The source of the close icons is an SVG so that could be used instead. I'll attach them here (as zip because github still doesn't support svg... :/).
[...]

PR #530, opened 2018-06-22
Add new close icons and make current icons Basilisk-specific

Lootyhoof:
Tag #317 

This `ifdef`'s Basilisk's icons in `global.css` and adds Pale Moon ones instead for generic platform styling (on Windows). A slightly different approach than what I'd originally planned, but hopefully a lot more minimal.

As a bonus, since Linux was also using different icons compared to previous Pale Moon releases, I've also done the same with these and just used the previous styling.

The styling itself works, but I've not been able to test the preprocessor side of things. This should probably be checked before merging.


trav90:
All seems good on Linux. :+1:

PR #549, opened 2018-06-26
Unify strings for "Restart in Safe Mode" - in Basilisk and Toolkit: "about:profiles" and "about:support"

janekptacijarabaci:
Ad
https://github.com/MoonchildProductions/Pale-Moon/issues/1563 - Rename "Restart with addons disabled" to "Restart in Safe Mode"
and
https://forum.palemoon.org/viewtopic.php?f=63&t=19506


janekptacijarabaci:
@JustOff 
FYI, `String changes`. Thanks.

PR #556, opened 2018-06-27
[PALEMOON] Use Promises.jsm instead of promise.js from jetpack in sanitize.js

### LINKED ISSUE #521: Stop using old-style promises.
We should convert any old-style promises in use to the final-spec internal promises implementation.

@JustOff has offered to tackle this (IIUC) so I'll assign him to this.

JustOff:
Get rid of "old-style" `promise.js` from the jetpack and replace it by `Promises.jsm` from the toolkit.

This resolves #521.


JustOff:
Wait a minute, I found extra copy of `promise.js` that needs to be removed. I'll do it after quick check.

JustOff:
Now it's ready.

PR #558, opened 2018-06-27
[PALEMOON] Include devtools/client when building language packs

### LINKED ISSUE #512: Pale Moon UXP language packs do not include devtools client
At the moment Palemoon UXP language packs do not include devtools `client` folder, but only `shared` at `\browser\chrome\en-US\locale\en-US\devtools\`, so we can't localize devtools properly.

I can manually add the necessary changes to the langpacks assembly, but I need help to update the build system to include devtools `client` when running `mach package`.

JustOff:
Based on [bug 1182722](https://bugzilla.mozilla.org/show_bug.cgi?id=1182722).

This resolves #512.


mattatobin:
Better ifdef MOZ_DEVTOOLS

PR #561, opened 2018-06-28
[PALEMOON] Use "w.gBrowser.selectedBrowser" instead of "aCurrentBrowser"

janekptacijarabaci:
Follow up #525 

__Steps to reproduce:__
Open browser, go to `Search bar`, click on `Search`

Throws an error in Browser Console:
```
ReferenceError: aCurrentBrowser is not defined  utilityOverlay.js:345:
```

Please review (@JustOff , @wolfbeast).

See also (creating a variable named `aCurrentBrowser`):
https://bugzilla.mozilla.org/show_bug.cgi?id=1000458
(`enter in location bar (URL bar) for bookmark keyword immediately followed by opening new tab sometimes loads url in new tab instead of old, particularly under high load`)

### DIFF
--- a/application/palemoon/base/content/utilityOverlay.js
+++ b/application/palemoon/base/content/utilityOverlay.js
@@ -347,7 +347,7 @@ function openLinkIn(url, where, params) {
                                 referrerPolicy: aReferrerPolicy,
                                 postData: aPostData,
                                 }); 
-    browserUsedForLoad = aCurrentBrowser;
+    browserUsedForLoad = w.gBrowser.selectedBrowser;
     break;
   case "tabshifted":
     loadInBackground = !loadInBackground;


JustOff:
LGTM.

wolfbeast:
This will fix the immediate problem but the issue is actually that in cadeea97bbefe42f89d16a4678fdc382c7b74bcd I forgot to define aCurrentBrowser which may also be passed through params to the `openLinkIn()` function.

So instead of this, we need:
```
  // We can only do this after we're sure of what |w| will be the rest of this function.
  // Note that if |w| is null we might have no current browser (we'll open a new window).
  var aCurrentBrowser = params.currentBrowser || (w && w.gBrowser.selectedBrowser);
```
below `var w = getTopWin();` and its sanity check.

PR #567, opened 2018-06-29
Undo Mozilla Bug 1302937; Reinstate -moz-mac-lion-theme media query

SpockMan02:
I have confirmed that it works as it did before it was removed.


mattatobin:
Looks alright to me judging from surrounding code. However, that is the depth of my impression.

+1 for whitespace cleanup along the way

PR #606, opened 2018-07-07
Updates to Reader and Narrator components.

Ascrod:
This patch fixes several issues with Reader and Narrator. Tag [#361](https://github.com/MoonchildProductions/UXP/issues/361).

- Remove telemetry from NarrateControls that is causing breakage.
- Fix local link handling in Reader Mode.
- Apply patch from bug [1456652](https://bugzilla.mozilla.org/show_bug.cgi?id=1456652). Note that this change does not resolve the issue in UXP - cookies are still being sent, as I mentioned [here](https://github.com/MoonchildProductions/UXP/issues/361#issuecomment-402850284).

### DIFF
--- a/toolkit/components/narrate/NarrateControls.jsm
+++ b/toolkit/components/narrate/NarrateControls.jsm
@@ -10,7 +10,6 @@ Cu.import("resource://gre/modules/narrate/VoiceSelect.jsm");
 Cu.import("resource://gre/modules/narrate/Narrator.jsm");
 Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/AsyncPrefs.jsm");
-Cu.import("resource://gre/modules/TelemetryStopwatch.jsm");
 
 this.EXPORTED_SYMBOLS = ["NarrateControls"];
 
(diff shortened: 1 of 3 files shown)


wolfbeast:
Please back out the change from 1456652 and I'll merge. if it has no effect then we shouldn't adopt the code. 

Ultimately, it's a matter of interpretation and not actually a security issue, because (quoted from BZ):

> Well... this is an interesting question. It amounts to whether the navigation from read:foo.com/site to foo.com/site is same-site or not. I think you could reasonably argue that it's same-site, but even if you don't think so, that this isn't a security issue as written.

> My reasoning for the latter is as follows. Suppose you had a site (V) using same-site: strict for cookies, and site (A) tried to exploit a CSRF issue in (V) by sending people to a deep-link in (V). same-site cookies would stop this, because the deep-linked website wouldn't have cookies.

> If (V) doesn't navigate when accessed without cookies (e.g. to an "authentication denied" or "something went wrong" style page with a different URL), then the URL in the URL bar would be the one (A) wanted it to be, right? But then, if the user went to the URL bar and hit enter to try to access the page again (because it "didn't work", whatever (A) promised would happen on (V) if they were trying to social engineer like you are here) then (V) *would* be exploited, because at this point it's just a user-requested access of the victim page, and same-site cookies won't help anymore (I'm assuming everyone agrees this is effectively by design, even if it may be considered unfortunate).

> In the same vein, why is exiting reader mode here different from the user requesting the URL directly? And in particular, given that (A) has no control over the content of (V), how would you encourage the user to either hit 'enter' in the URL bar, or click "exit reader mode", rather than closing the tab (which after all is newly opened in this exploit...)?

Ascrod:
Understood. Patch reverted.

PR #609, opened 2018-07-07
Fix typos in comments - memory.properties

janekptacijarabaci:
Align `Name values` with `names in comments`.

@JustOff - FYI (for a context).


JustOff:
Pushed to CrowdIn.

PR #619, opened 2018-07-10
Pale Moon: Fix positioning of caption buttons in application-drawn ti…

### LINKED ISSUE #586: Pale Moon: macOS: Caption buttons placed incorrectly in application-drawn titlebar
Now that the titlebar is back with lightweight themes, and private browsing windows have caption buttons (#585, #588)...

screenshot 2018-07-03 01 53 23

Correct positioning is on the left of the titlebar, not the top-right. I've been able to move them to the bottom-right with CSS, but that's still wrong and I haven't figured out how to fix it properly.

SpockMan02:
…tlebar

This resolves #586.

### DIFF
--- a/application/palemoon/base/content/browser.xul
+++ b/application/palemoon/base/content/browser.xul
@@ -334,7 +334,9 @@
       </button>
     </hbox>
 #endif
+#ifndef XP_MACOSX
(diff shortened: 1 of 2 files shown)


wolfbeast:
Changes browser XUL -- might need theme changes?

PR #661, opened 2018-07-21
[PALEMOON] Places - remove unused @param from controller.js

janekptacijarabaci:
`_hasRemovableSelection()` does not use parameters.


wolfbeast:
Just a comment update...

PR #694, opened 2018-08-04
Fix constructor in nsHtml5AtomTable.cpp

adeshkp:
Follow up to 2bc00ef to fix build bustage


wolfbeast:
Sorry, I backed out your previous patch for build bustage.

wolfbeast:
there- re-applied and applied this follow-up. please check if it all looks as intended.

adeshkp:
All good now!

PR #704, opened 2018-08-11
Shell quote environment variable values when dumping them in dump_env.py

trav90:
The mozconfig output parsing code already (mostly) handles shell quoted strings, because that's what `set` outputs. By quoting environment variable values, we avoid a bunch of problems with "weird" values.

On some newer distros (Fedora 28 being the primary culprit here) all mach commands fail because the environment variables are incorrectly parsed. This patch changes how the build system takes values from the environment to feed build options to configure (there is no affect on the application being built).

Tested and verified that the build completes successfully on Fedora 28 without busting building on CentOS 7.

### DIFF
--- a/python/mozbuild/mozbuild/action/dump_env.py
+++ b/python/mozbuild/mozbuild/action/dump_env.py
@@ -6,5 +6,11 @@
 # native paths printed on Windows so that these paths can be incorporated
 # into Python configure's environment.
 import os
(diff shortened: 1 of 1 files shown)


khronosschoty:
built without issues with this patch.

trav90:
Thanks @khronosschoty!

Issue #705, opened 2018-08-11
Support GCC 8


trav90:
While UXP can be compiled with GCC 8 out of the box, there's a lot of warnings spewed by the compiler that need to be investigated before we can say we officially support GCC 8.

kn-yami:
FYI an user at #palemoon reported that building Pale Moon with GCC 8.2 on Gentoo does not work.

khronosschoty:
Can't reproduce this on Slackware.

PR #706, opened 2018-08-11
Fix/silence some warnings with GCC 8

trav90:
This PR fixes and/or silences several compiler warnings output by GCC 8. With this patch set, about 35 compiler warnings are fixed or silenced when building Pale Moon.

In addition to testing GCC 8 I also built Basilisk using GCC 7 and Pale Moon using GCC 7 and 4.9. I saw no issues in either the build process or the resulting binary using any of these compiler versions.

Tag #705


khronosschoty:
I've been building with GCC 8 series now for a while and no issues. FWIW

PR #708, opened 2018-08-15
fix basilisk mac icons, executable name, and min os version

wicknix:


wolfbeast:
I didn't know there were plans to actually publish Basilisk for Mac. It never crossed _my_ mind at least.

mattatobin:
The plist file is preprocessed shouldn't it use `MOZ_APP_NAME`. Aside from that even if we aren't going to publish Basilisk for Macintosh right away it still needs done.

wolfbeast:
@mattatobin I'm not sure if I ever want to publish official builds of Basilisk on Macintosh. But sure, getting the assets in can't hurt anything, I guess.

Issue #457, opened 2018-06-06
Investigate and solve compiler warnings.


wolfbeast:
For as far as our own (not third-party) code is concerned, we should investigate and where possible solve compiler warnings emitted by the build process.

PR #718, opened 2018-08-20
Remove unused variable

trav90:
Just what it says in the tin. Followup to commit fc001cc.

Fixes a compiler warning with all supported versions of GCC.


wolfbeast:
Tag #457

PR #720, opened 2018-08-21
Update ShouldAddToSessionHistory() rules for "about:newtab" and "about:logopage"

### LINKED ISSUE #719: Tab history does not store the "about:newtab" page, but the "about:logopage" is stored
As reported on the forum, [users can no longer go back in history to QuickDial](https://forum.palemoon.org/viewtopic.php?f=3&t=20024) (`about:newtab`). At the same time, the `about:logopage` is now stored in the tab history, while it should not.

JustOff:
This excludes the `about:logopage` from the session history and allows the `about:newtab` to be stored based on `browser.newtabpage.add_to_session_history`.

This resolves #719.

### DIFF
--- a/docshell/base/nsDocShell.cpp
+++ b/docshell/base/nsDocShell.cpp
@@ -12281,11 +12281,9 @@ nsDocShell::ShouldAddToSessionHistory(nsIURI* aURI)
       return false;
     }
 
-    if (buf.EqualsLiteral("blank") || buf.EqualsLiteral("logopage")
-#ifndef MC_PALEMOON
-        || buf.EqualsLiteral("newtab")
-#endif
-       ) {
+    if (buf.EqualsLiteral("blank") || buf.EqualsLiteral("logopage") ||
+        buf.EqualsLiteral("newtab") &&
+        !Preferences::GetBool("browser.newtabpage.add_to_session_history", false)) {
       return false;
     }
   }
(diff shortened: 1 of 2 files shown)


docshell/base/nsDocShell.cpp
```diff
@@ -12281,7 +12281,11 @@ nsDocShell::ShouldAddToSessionHistory(nsIURI* aURI)
       return false;
     }
 
-    if (buf.EqualsLiteral("blank") || buf.EqualsLiteral("newtab")) {
+    if (buf.EqualsLiteral("blank") || buf.EqualsLiteral("logopage")
+#ifndef MC_PALEMOON
```
wolfbeast: I think this can be done platform-wide. Is there a reason for applications other than Pale Moon to _not_ have a history entry for `about:newtab`?
What do you think, @mattatobin @g4jc ?

mattatobin: I don't think any of those internal pages should have a history entry. Be it blank, newtab, or logopage.

JustOff: AFAIR, `about:newtab` was always stored in the history until Mozilla [screwed it up](https://bugzilla.mozilla.org/show_bug.cgi?id=724239) in FF15. However, PM retained the old behavior up to v27, and everyone was happy. I guess this was just unintentionally changed in v28 due to the platform update, and it's worth to be fixed at least for PM, if not platform-wide.

g4jc: @wolfbeast I would prefer not storing history of internal pages for privacy reasons. I can't think of any reason why their history would needed for users. It's fine to keep this a local ifdef though.

wolfbeast: @g4jc we're discussing this in the attached issue #719 and have settled on putting this behind a pref (it's a sec issue, also) disabled by default, which would allow you to configure it as you want.

PR #729, opened 2018-08-24
Change mozilla.pdf to UXP.pdf

zqqw:
Regression of Tycho behaviour:
https://github.com/MoonchildProductions/Pale-Moon/commits/6c1009bbf22aff8a003d296de1781126eb6906a7/widget/gtk/nsDeviceContextSpecG.cpp

Print to pdf or ps uses default file name of mozilla.p*
UXP.p* works in both Pale Moon and Basilisk, and is shorter, reducing binary and codebase size, improving performance as well as aesthetics.

### DIFF
--- a/widget/gtk/nsDeviceContextSpecG.cpp
+++ b/widget/gtk/nsDeviceContextSpecG.cpp
@@ -415,9 +415,9 @@ NS_IMETHODIMP nsPrinterEnumeratorGTK::InitPrintSettingsFromPrinter(const char16_
     path = PR_GetEnv("HOME");
   
   if (path)
-    filename = nsPrintfCString("%s/mozilla.pdf", path);
+    filename = nsPrintfCString("%s/UXP.pdf", path);
   else
-    filename.AssignLiteral("mozilla.pdf");
+    filename.AssignLiteral("UXP.pdf");
 
   DO_PR_DEBUG_LOG(("Setting default filename to '%s'\n", filename.get()));
   aPrintSettings->SetToFileName(NS_ConvertUTF8toUTF16(filename).get());


wolfbeast:
Perhaps it would be better to simply use "document.pdf", super-generic.
The few bytes binary and code base size difference because of this is really not an argument to consider, and it will most definitely NOT impact performance :)

zqqw:
UXP.pdf (or goanna.pdf) is less likely to conflict with any existing files, compared to document.pdf, also it gives some clue as to the origin of the file if the user forgot what it was.
While insignificant, those few bytes might be better used elsewhere - and where did I read this?
Improving Mozilla code: You know you're on the right track with code changes when you spend the majority of your time deleting code.

EbonJaeger:
"You know you're on the right track with code changes when you spend the majority of your time deleting code"
There is such a thing as taking it too far. :)

As for making it more clear where the file came from, any default name will be generic enough to where you can't tell anyway. I'd even argue that calling it UXP.pdf is more confusing, because then it actually looks like it might be for something else.

zqqw:
OK, so what do you all want to call it? Or are you happy to leave it as mozilla? It's a pretty insignificant change, but was apparently thought worthwhile before.

wolfbeast:
I honestly don't think too much time should be spent on choosing a name. The problem I have with "UXP.pdf" or any other `technology name` used is that it's easy to confuse with "this document is about `technology name`" -- in that respect palemoon.pdf and goanna.pdf were also not good names, neither is mozilla.pdf.

`document.pdf` may conflict with existing files, but so would any other name if you choose to save to the same location. But at least it won't imply anything about the content -- file names usually hint at their contents not their origin. That's my main reasoning behind it.

Makes sense? Yes? No? Pick something generic or leave it as-is, I guess.

mattatobin:
document-yymmddhhmmss.pdf is my suggestion.

zqqw:
> document-yymmddhhmmss.pdf is my suggestion
I like that idea - then you can leave it as default yet save multiple files. I'll see if I can do that and test it.

wolfbeast:
Closing this issue as it needs more code work for the datestamp.

Please file a new PR when you have something to merge, linked to the issue #744 created to track since it's no longer a trivial codegraft from previous work.

PR #747, opened 2018-09-03
Update 'browser.newtab.url' when it's used for new tabs and 'browser.startup.homepage' changes

### LINKED ISSUE #746: Sometimes the "when opening a new tab show = my home page" setting in preferences/tabs will not pick up a home page change
As reported on the [forum](https://forum.palemoon.org/viewtopic.php?f=29&t=20187):

> Actually, the sequence where it always seems to occur is (with new, clean profile):
> 
> 1). Set home page and then go to preferences/tabs/set new tab as your home page, press ok. It works good. Home page and new tab agree.
> 2). Change home page.. new tab reflects new/current home page change correctly - home page and new tab agree - it obeys the new tab set as your home page.
> 3). Change home page again.. new tab reflects old (last) home page and doesn't show current home page. Home page and new tab do not agree. I have go to preferences/tabs, and press OK again to make it show current home page.

JustOff:
Update `writeNewtabUrl()`, move it from `tabs.js` to a separate module and call it when necessary.

This resolves #746.

### DIFF
--- a/application/palemoon/components/preferences/newtaburl.js
+++ b/application/palemoon/components/preferences/newtaburl.js
@@ -1,0 +1,66 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var gNewtabUrl = {
+  /**
+   * Writes browser.newtab.url with the appropriate value.
+   * If the choice is "my home page", get and sanitize
+   * the browser home page URL to make it suitable for newtab use.
+   *
+   * Called from prefwindow ondialogaccept in preferences.xul,
+   * newtabPage oncommand in tabs.xul, browserHomePage oninput,
+   * useCurrent, useBookmark and restoreDefaultHomePage oncommand
+   * in main.xul to consider instantApply.
+   */
+  writeNewtabUrl: function(newtabUrlChoice, browserHomepageUrl) {
+    try {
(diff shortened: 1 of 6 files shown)


application/palemoon/components/preferences/newtaburl.js
```diff
@@ -0,0 +1,66 @@
...
+        browserHomepageUrl = Services.prefs.getComplexValue("browser.startup.homepage",
+                              Components.interfaces.nsIPrefLocalizedString).data;
+      }
+      let newtabUrlPref = Services.prefs.getCharPref("browser.newtab.url");
+      switch (newtabUrlChoice) {
+        case 1:
+          newtabUrlPref = "about:logopage";
+          break;
+        case 2:
+          newtabUrlPref = "http://start.palemoon.org/";
```
mattatobin: If you are gonna go to the trouble of modifying this stuff it would be nice if you could use the default version of the preference for the homepage for this case instead of a hardcoded url.

JustOff: That's how it was [initially created](https://github.com/MoonchildProductions/Pale-Moon/commit/aff0e2bf91fb425ee8ecaaf956b9c7d2fef67e5b#diff-c428896aeab9ab43c4c41c32ca5ea4ceR114) by @wolfbeast; we can improve it by a separate task if necessary.

wolfbeast: Just do it here instead of a separate task.
it should be easy with `getDefaultPref()` (IIRC) to get the default pref for the home page.

PR #754, opened 2018-09-07
Fix more -Wclass-memaccess warnings (GCC8)

trav90:
This PR fixes another handful of compiler warnings when building with GCC 8. These are similar fixes as what was done in #714 (see the individual commits for more specifics).

Built and tested using GCC 4.9, GCC 7.3 and GCC 8 with no problems seen in any of the builds after spending a fair amount of time browsing the net. Requesting review (due to my unfamiliarity with code in js/).

Tag #705 and #457

### DIFF
--- a/js/src/ds/LifoAlloc.h
+++ b/js/src/ds/LifoAlloc.h
@@ -15,6 +15,8 @@
 #include "mozilla/TemplateLib.h"
 #include "mozilla/TypeTraits.h"
 
+#include <new>
+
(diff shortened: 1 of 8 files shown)


wolfbeast:
I'm not sure if I can properly review this without extensive research of the surrounding code, myself...

wolfbeast:
Been giving this a thorough workout with some heavy benchmarks and general use without issues, so Windows seems to be fine with these changes also.

PR #767, opened 2018-09-13
Fix more -Wclass-memaccess warnings (part 2)

trav90:
More fixes in the same vein as the ones in #754. This patch set fixes around a dozen compiler warnings with GCC 8.

Tested with GCC 7 & 8, with no issues or problems seen in either build.

Tage #457


wolfbeast:
Built Basilisk-x86 with the result - ran a few JS benchmarks and did not see any issues or regressions.

PR #774, opened 2018-09-17
Use SessionStore.promiseInitialized() to avoid race condition in "about:home"

### LINKED ISSUE #772: "Restore Previous Session" is missing from Pale Moon's Start Page
This is happening at least in Pale Moon 28.0.1, 64b, i'm not aware for other versions. 

Steps to reproduce: 
1. In Preferences -> General, leave "Show my home page" (default). 
    In "Home Page:", delete "`http://start.palemoon.org/`". 
2. Close the browser with at least two tabs and then open it again. 

When Pale Moon starts, it has one tab with its home page, but the option 
"Restore Previous Session" is missing. 
Investigating this issue, i saw that if the home page is refreshed once, the 
missing option gets appeared. 

I think this issue must be prioritized because gives to the user the impression 
that there is no session to be restored.

JustOff:
This ensures that SessionStore is initialized before checking whether the last session can be restored.

Backported from Basilisk, resolves #772.

### DIFF
--- a/application/palemoon/base/content/browser.js
+++ b/application/palemoon/base/content/browser.js
@@ -2423,11 +2423,12 @@ function BrowserOnAboutPageLoad(doc) {
   if (doc.documentURI.toLowerCase() == "about:home") {
     if (!PrivateBrowsingUtils.isWindowPrivate(window)) {
       let wrapper = {};
-      Components.utils.import("resource:///modules/sessionstore/SessionStore.jsm", wrapper);
+      Cu.import("resource:///modules/sessionstore/SessionStore.jsm", wrapper);
       let ss = wrapper.SessionStore;
       ss.promiseInitialized.then(function() {
-        if (ss.canRestoreLastSession)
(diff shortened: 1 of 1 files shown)


JustOff:
I think it's worth it to be included in the 28.1.0 release.

JustOff:
I wonder what happened with this PR, I've never seen anything like this:

 * wolfbeast merged **2 commits** into MoonchildProductions:master from JustOff:PR_about_home_session_restore
* There is no merge commit in the master tree
* In fact, only the first commit was merged
* The merge looks like a cherry-pick

@wolfbeast, is everything ok here?

wolfbeast:
It was squash merged. The second commit was merged as well -- it was combined into a single code commit.

JustOff:
Ok, thanks, it seems I have to learn more about git :thinking:

wolfbeast:
I usually don't squash -- just with the inconsequential style commit it was rather pointless to have a separate commit for that

PR #778, opened 2018-09-19
Fix timer overflow on converting from sec to msec in idleService

### LINKED ISSUE #723: High CPU usage on 32bit platforms when left idle
Creating this issue to track high CPU Usage on 32bit platforms. The bug is not reproducible on 64bit.
So far this bug has shown itself in Iceweasel-UXP and PaleMoon when running on 32bit.

I was able to reproduce this on a Linux VM with a 32bit distro set to a single CPU, and 1GB of ram.
I believe it to be the same bug affecting [Windows users of PM28](https://forum.palemoon.org/viewtopic.php?f=29&p=148995).

1. Start the browser application
2. Open a new tab and get the browser out of focus by using any other program and not clicking on the browser for 1 minute.
3. CPU usage will climb quickly until reaching and hanging at 100%

So far gdb has resulted in very little information other than it is hanging during an nsThread call. I have been unable to locate which call and had originally suspected a runaway Telemetry timer due to the requirement of waiting one minute prior to the hang. However,  removing those timers has had no affect.

Using the [Built-In Profiler](https://developer.mozilla.org/en-US/docs/Mozilla/Performance/Profiling_with_the_Built-in_Profiler) may help here. More research is needed.

MihailZenkov:
In some cases idleService set mDeltaToNextIdleSwitchInS to UINT32_MAX than convert it to msec (double, UINT32_MAX*1000) and try init nsTimerImpl with this value. But nsTimerImpl limited to uint32_t so we got overflow on conversion from double to uint32_t.

GCC produce different behavior for same code on different arch (in our case 64bit and 32bit).

If we convert double bigger than UINT32_MAX to uint32_t on 64bit - we have some big value. But if we do same on 32bit - we have zero.

So we got incorrect timer value in both arch but on 32bit we got timer restart with zero delay and have high CPU usage.

Resolves #723.

### DIFF
--- a/widget/nsIdleService.cpp
+++ b/widget/nsIdleService.cpp
@@ -46,6 +46,10 @@ using namespace mozilla;
 // Number of seconds in a day.
 #define SECONDS_PER_DAY 86400
 
+// MAX_DELTA_SEC * 1000 should be less than UINT32_MAX to prevent overflow on
+// converting from sec to msec
+#define MAX_DELTA_SEC (SECONDS_PER_DAY * 10)
+
 static PRLogModuleInfo *sLog = nullptr;
 
 #define LOG_TAG "GeckoIdleService"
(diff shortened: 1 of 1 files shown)


wolfbeast:
Nice catch!

PR #788, opened 2018-09-24
gfx: add pref for enabling APZ without e10s

### LINKED ISSUE #345: Allow APZ on desktop (experimental)
Mozilla has built in a hard-disable for APZ on desktop based on whether e10s is enabled or not.
We should allow APZ on desktop as an experimental feature.

https://github.com/MoonchildProductions/UXP/blob/0264de4c3b0f74fb35ac13f2726e7f926d705ea5/gfx/thebes/gfxPlatform.cpp#L2399-L2416

alaviss:
Resolves #345 

Tested on Linux with Basilisk. It might need a better pref name, but I this is the best I can came up with.

### DIFF
--- a/gfx/thebes/gfxPlatform.cpp
+++ b/gfx/thebes/gfxPlatform.cpp
@@ -2380,13 +2380,13 @@ gfxPlatform::AsyncPanZoomEnabled()
 {
 #if !defined(MOZ_WIDGET_ANDROID) && !defined(MOZ_WIDGET_UIKIT)
   // For XUL applications (everything but Firefox on Android) we only want
-  // to use APZ when E10S is enabled. If we ever get input events off the
-  // main thread we can consider relaxing this requirement.
-  if (!BrowserTabsRemoteAutostart()) {
+  // to use APZ when E10S is enabled or when the user explicitly enable it.
+  if (BrowserTabsRemoteAutostart() || gfxPrefs::APZDesktopEnabled()) {
+    return gfxPrefs::AsyncPanZoomEnabledDoNotUseDirectly();
+  } else {
     return false;
   }
-#endif
-#ifdef MOZ_WIDGET_ANDROID
+#elif defined(MOZ_WIDGET_ANDROID)
   return true;
 #else
   return gfxPrefs::AsyncPanZoomEnabledDoNotUseDirectly();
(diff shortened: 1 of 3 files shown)


gfx/thebes/gfxPlatform.cpp
```diff
@@ -393,6 +393,8 @@ NS_IMPL_ISUPPORTS(SRGBOverrideObserver, nsIObserver, nsISupportsWeakReference)
 
 #define GFX_PREF_CMS_FORCE_SRGB "gfx.color_management.force_srgb"
 
+#define GFX_PREF_EXPERIMENTAL_APZ "gfx.experimental.apz"
```
wolfbeast: It would be better/more logical to make this part of the `apz.*` pref tree instead of `gfx.*`. I'd also like to avoid "experimental" in pref names -- what is experimental now might not be later on and then you'd have to rename the pref again.

So, perhaps something like `apz.desktop.enabled`?

gfx/thebes/gfxPlatform.cpp
```diff
@@ -2382,7 +2384,8 @@ gfxPlatform::AsyncPanZoomEnabled()
   // For XUL applications (everything but Firefox on Android) we only want
   // to use APZ when E10S is enabled. If we ever get input events off the
```
wolfbeast: Please change the associated comment here to reflect the changes with pref control.

gfx/thebes/gfxPlatform.cpp
```diff
@@ -2382,7 +2384,8 @@ gfxPlatform::AsyncPanZoomEnabled()
...
   // main thread we can consider relaxing this requirement.
-  if (!BrowserTabsRemoteAutostart()) {
+  if (!BrowserTabsRemoteAutostart() &&
+      !Preferences::GetBool(GFX_PREF_EXPERIMENTAL_APZ, false)) {
```
wolfbeast: It's probably better to rewrite this little bit of logic to make it a positive check (if e10s or pref then use APZ). Also, check and see what `AsyncPanZoomEnabledDoNotUseDirectly()` does to limit further use of APZ? What extra conditions does it apply?

alaviss: It links to this pref: https://github.com/MoonchildProductions/UXP/blob/dc9d933d81646b53edab245eb8f96b1e82ba752f/gfx/thebes/gfxPrefs.h#L457

alaviss:
@wolfbeast I've made all the changes you requested, please review.

PR #789, opened 2018-09-26
backport mozbug 1444532 - fix a leak in SHA256 in nsHttpConnectionInfo.cpp r=mayhemer

g4jc:
The original code (from bug 1200802) declared an XPCOM object as a static bare pointer,
which for future reference is probably never the right thing to do. It might have worked if
it was cleared before shutdown but it never was.


g4jc:
@wolfbeast it randomly closed when I "push --forced" that is new to me, still investigating how to re-open it. Github is displaying "Pull request successfully merged and closed" here, which is false.

g4jc:
Alright, spaces are (2) and I also bumped the "hasher = do_CreateInstance" into one line for clarity.

PR #790, opened 2018-09-26
backport mozbug 1344613 - Avoid possibility of null pointer crash in nsSOCKSIOLayer.cpp r=mayhemer

g4jc:
This fixes a crash reported while using socks5.


wolfbeast:
I'm pretty sure we're using our zeroing allocator for this as opposed to Mozilla, but it can't hurt to be explicit.

PR #791, opened 2018-09-26
Issue #792 - backport mozbug 1334776 - CVE-2017-7797 Header name interning leaks across origins

### LINKED ISSUE #792: CVE-2017-7797 Header name interning leaks across origins
Per the [CVE](https://nvd.nist.gov/vuln/detail/CVE-2017-7797): _Response header name interning does not have same-origin protections and these headers are stored in a global registry. This allows stored header names to be available cross-origin. This vulnerability affects Firefox < 55._

----

Potential attack: session super-cookie.

[Moz Notes](https://bugzilla.mozilla.org/show_bug.cgi?id=1334776#c5):
"The problem is that for unknown header names we store the first one we see and then later we case-insensitively match against that name *globally*. That means you can track if a user agent has already seen a certain header name used (by using a different casing and observing whether it gets normalized). This would allow you to see if a user has used a sensitive service that uses custom header names, or allows you to track a user across sites, by teaching the browser about a certain header case once and then observing if different casings get normalized to that.

What we should do instead is only store the casing for a header name for each header list and not globally. That way it only leaks where it's expected (and necessary) to leak."

[Moz fix note](https://bugzilla.mozilla.org/show_bug.cgi?id=1334776#c8):
"nsHttpAtom now holds the old nsHttpAtom and a string that is case sensitive (only for not standard headers).
So nsHttpAtom holds a pointer to a header name. (header names are store on a static structure). This is how it used to be. I left that part the same but added a nsCString which holds a string that was used to resoled the header name. So when we parse headers we call ResolveHeader with a char*. If it is a new header name the char* will be stored in a HttpHeapAtom, nsHttpAtom::_val will point to HttpHeapAtom::value and the same strings will be stored in mLocalCaseSensitiveHeader. For the first resolve request they will be the same but for the following maybe not. At the end this nsHttpAtom will be stored in nsHttpHeaderArray. For all operation we will used the old char* except when we are returning it to a script using VisitHeaders."

----

I've backported this in a PR and all seems to be working well, but as @wolfbeast mentioned, this does change how HTTP headers are processed. It also makes them lowercase. I am unsure if this affects addons such as Modify Headers or similar, but that is the only major compatibility issue I can think of at the moment.

g4jc:
Potential attack: session super-cookie.

[Moz Notes](https://bugzilla.mozilla.org/show_bug.cgi?id=1334776#c5):
"The problem is that for unknown header names we store the first one we see and then later we case-insensitively match against that name *globally*. That means you can track if a user agent has already seen a certain header name used (by using a different casing and observing whether it gets normalized). This would allow you to see if a user has used a sensitive service that uses custom header names, or allows you to track a user across sites, by teaching the browser about a certain header case once and then observing if different casings get normalized to that.

What we should do instead is only store the casing for a header name for each header list and not globally. That way it only leaks where it's expected (and necessary) to leak."

[Moz fix note](https://bugzilla.mozilla.org/show_bug.cgi?id=1334776#c8):
"nsHttpAtom now holds the old nsHttpAtom and a string that is case sensitive (only for not standard headers).
So nsHttpAtom holds a pointer to a header name. (header names are store on a static structure). This is how it used to be. I left that part the same but added a nsCString which holds a string that was used to resoled the header name. So when we parse headers we call ResolveHeader with a char*. If it is a new header name the char* will be stored in a HttpHeapAtom, nsHttpAtom::_val will point to HttpHeapAtom::value and the same strings will be stored in mLocalCaseSensitiveHeader. For the first resolve request they will be the same but for the following maybe not. At the end this nsHttpAtom will be stored in nsHttpHeaderArray. For all operation we will used the old char* except when we are returning it to a script using VisitHeaders."

----
Fixes  #792

### DIFF
--- a/devtools/client/netmonitor/filter-predicates.js
+++ b/devtools/client/netmonitor/filter-predicates.js
@@ -72,7 +72,7 @@ function isWS({ requestHeaders, responseHeaders }) {
 
   // Find the 'upgrade' header.
   let upgradeHeader = requestHeaders.headers.find(header => {
-    return (header.name == "Upgrade");
+    return (header.name.toLowerCase() == "upgrade");
   });
 
   // If no header found on request, check response - mainly to get
@@ -81,7 +81,7 @@ function isWS({ requestHeaders, responseHeaders }) {
   if (!upgradeHeader && responseHeaders &&
       Array.isArray(responseHeaders.headers)) {
     upgradeHeader = responseHeaders.headers.find(header => {
-      return (header.name == "Upgrade");
+      return (header.name.toLowerCase() == "upgrade");
     });
   }
 
--- a/devtools/client/netmonitor/test/browser_net_copy_headers.js
+++ b/devtools/client/netmonitor/test/browser_net_copy_headers.js
@@ -49,12 +49,12 @@ add_task(function* () {
 
   const EXPECTED_RESPONSE_HEADERS = [
     `${httpVersion} ${status} ${statusText}`,
-    "Last-Modified: Sun, 3 May 2015 11:11:11 GMT",
-    "Content-Type: text/html",
-    "Content-Length: 465",
-    "Connection: close",
-    "Server: httpd.js",
-    "Date: Sun, 3 May 2015 11:11:11 GMT"
+    "last-modified: Sun, 3 May 2015 11:11:11 GMT",
+    "content-type: text/html",
+    "content-length: 465",
+    "connection: close",
+    "server: httpd.js",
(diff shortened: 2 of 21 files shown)


wolfbeast:
It would be good in this case to actually create an issue with an explanation of the issue this PR resolves, with the appropriate flags. This significantly changes HTTP header behavior and I'd rather not have this only mentioned in a dangling PR, no matter the excellent PR description. Could you please do that?

wolfbeast:
Thanks for creating the issue. This resolves #792

PR #804, opened 2018-10-01
[PALEMOON] Use message manager to detect full-screen HTML5 video in S4E module

### LINKED ISSUE #803: Setting to hide the popup window in full-screen HTML5 video does not work in Pale Moon
[As reported on the forum](https://forum.palemoon.org/viewtopic.php?f=13&t=20577), Pale Moon does not hide the status text in full-screen HTML5 video when it's selected to display in a pop-up.

JustOff:
Since UXP loads documents asynchronously, we can no longer directly check `document.mozFullScreenElement` when detecting full-screen HTML5 video in S4E module.

This PR resolves this by using message manager, but only when `status4evar.advanced.status.detectFullScreen` is enabled (`false` by default).

Based on [S4E#30](https://github.com/SparkyBluefang/S4E/issues/30), resolves #803.

### DIFF
--- a/application/palemoon/components/statusbar/Status.jsm
+++ b/application/palemoon/components/statusbar/Status.jsm
@@ -35,7 +35,7 @@ S4EStatusService.prototype =
   _defaultStatus:          { val: "", type: "" },
 
   _isFullScreen:           false,
-  _isFullScreenVideo:      false,
+  _isVideo:                false,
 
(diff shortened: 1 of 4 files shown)


mattatobin:
Looks good on paper.. If it works then r+

JustOff:
I verified that it works as expected using a private build.

PR #801, opened 2018-09-30
Update ffvpx code to 4.0.2

trav90:
This PR updates our ffvpx code to 4.0.2 (FFmpeg's most recent release).

I've tested VP9 videos on YouTube, in addition to the vpx files here:

http://base-n.de/webm/VP9%20Sample.html

No issues seen on Linux; verification needed for Windows.

Issue #83, opened 2018-03-20
Keep FFmpeg code in sync with upstream


trav90:
_Clone of https://github.com/MoonchildProductions/moebius/issues/24_

UXP bundles FFmpeg code to use for vp8/9 playback on x86 platforms (see http://xref.palemoon.org/uxp-trunk/source/media/ffvpx/README_MOZILLA). To ensure that UXP applications ship with all relevant bug fixes and security patches from FFmpeg this code should periodically be re-synced with upstream.

trav90:
As of c75dae3ed21bfa5a8ae46cd83d18329af5bea05a our code is synced with FFmpeg 3.4.2.

mattatobin:
So shouldn't we stay in sync with upstream?

wolfbeast:
Well Travis did the media stuff; he's assigned here. *I* Don't have the free time for it. Anyone else lacking his presence?

PR #806, opened 2018-10-01
Update ffvpx code to 4.0.2 (take 2)

trav90:
2nd attempt at updating our ffvpx code to the latest release.

Same as #801, with some symbols and dummy functions cleaned up which should resolve the build bustage on Windows (these symbols and functions are optimized out by GCC hence why the build bustage didn't affect Linux).

Tag #83


wolfbeast:
Clobber build successful and tested.
image1

PR #811, opened 2018-10-04
Fix in-tab close button behavior

### LINKED ISSUE #775: Quickly pressing the close button on tabs has a delay which is not seen in Firefox or chrome
new profile

Open 10-20 tabs

now quickly press close button/x button on tabs to close 5-10 tabs,
tabs close but when closing quickly it has a wait of 1second or so which ignores next click to close.
which causes another click required to close. Basically 2 clicks to close as opposed to one click when closing tabs fast(less than half a second).

when closing these tabs quickly Double clicking the x button closes the tabs instantly no matter how fast.

now try closing the tabs using x button slowly(1sec between clicks)
tabs close with one click instead of two.

In Firefox and chrome it's not an issue no matter how fast the user closes the tabs.

KB shortcut works fine.

can you guys fix this quickly

@JustOff 
@wolfbeast 

win/lin

MaxKoll:
- Fixes #775.
- Coincidentally also fixes a regression that prevented opening new tabs via double click on an empty space in the tab bar.

### DIFF
--- a/application/palemoon/base/content/tabbrowser.xml
+++ b/application/palemoon/base/content/tabbrowser.xml
@@ -4296,17 +4296,94 @@
             event.originalTarget.localName != "box")
           return;
 
-        // See hack note in the tabbrowser-close-tab-button binding
+        // See comments in the "mousedown" and "click" event handlers of the
+        // "tabbrowser-tabs" binding.
         if (!this._blockDblClick)
           BrowserOpenTab();
+        else
+          this.flagBlockedOpenTabAfterDblClick = true;
 
         event.preventDefault();
       ]]></handler>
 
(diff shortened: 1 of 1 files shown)


MaxKoll:
I realize now that there are still problems when tabs are animated. Unfortunately I did not consider animated tabs before.

Also there was no regression as mentioned in the commit message: I tested against my daily-use stable build that has tabs above the url bar. Here, double clicking the tab bar never opens new tabs (as intended), which I misinterpreted.

I'm seeing a revert button here that I'd like to try out soon, and I'm going push a revised commit afterwards. Hopefully this is the preferable course of action, otherwise I'd be thankful for a hint.

PR #813, opened 2018-10-06
[BASILISK] Protect against nsIPrincipal.origin throwing for about:blank iframes and custom protocol urls

JustOff:
This protects `SessionStorage.jsm` against `nsIPrincipal.origin` throwing; based on [bug 1344595](https://bugzilla.mozilla.org/show_bug.cgi?id=1344595).

Currently it happens with custom protocol urls (for example `caa:` from [Classic Add-ons Archive](https://github.com/JustOff/ca-archive)), but will also apply if we decide to implement [bug 1340710](https://bugzilla.mozilla.org/show_bug.cgi?id=1340710).

The issue was not created as it's a small minor one-file change.

### DIFF
--- a/application/basilisk/components/sessionstore/SessionStorage.jsm
+++ b/application/basilisk/components/sessionstore/SessionStorage.jsm
@@ -74,7 +74,14 @@ var SessionStorageInternal = {
 
       // Get the origin of the current history entry
       // and use that as a key for the per-principal storage data.
-      let origin = principal.origin;
+      let origin;
+      try {
+        // The origin getter may throw for about:blank iframes as of bug 1340710,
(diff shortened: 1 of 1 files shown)


wolfbeast:
FTR, Bug 1340710 = Speed up principal equality and subsumption checks. Which is honestly quite silly to have that much code churn for, and has anyone actually checked if this would ever be even noticeable?

JustOff:
@g4jc, you might also want to have this applied to Iceweasel.

---

> Bug 1340710 = Speed up principal equality and subsumption checks [..] has anyone actually checked if this would ever be even noticeable?

In theory, this may be so, but I didn't see any mention of any real benchmarks there.

Issue #87, opened 2018-03-29
Support AV1 video


wolfbeast:
The groundwork for this was done by Mozilla in https://bugzilla.mozilla.org/show_bug.cgi?id=1314147 as the first browser to get a reference implementation in.
Since the distance from our fork point is relatively small yet (FF55), it _should_ be fairly straightforward to port this across to UXP.
As the codec is still being finalized, there might need to be some adjustments made after this bug's set of patches and related bugs mentioned/linked, and it's imperative that this remains as experimental (and disabled by default) for the time being, but getting this work in now will prevent blocking issues later on.

g4jc:
Related bug: aom update -> https://bugzilla.mozilla.org/show_bug.cgi?id=1445683

PR #818, opened 2018-10-07
Add mozilla::Span

trav90:
mozilla:Span is used in multiple places in Mozilla's AOM implementation for AV1 support. Adding mozilla:Span to our tree will make it much easier for us to add AV1 codec support in the future (#87).

Since MFBT and C++ templates are not my strong points I'm requesting a review here. The new Span.h header includes lots of comments and documentation, and here are a couple additional links:

Mozilla's implementation: [BMO 1295611](https://bugzilla.mozilla.org/show_bug.cgi?id=1295611)
Portions of Microsoft's GSL library that Mozilla's implementation is based on: [here](https://github.com/Microsoft/GSL/blob/3819df6e378ffccf0e29465afe99c3b324c2aa70/include/gsl/span) and [here](https://github.com/Microsoft/GSL/blob/3819df6e378ffccf0e29465afe99c3b324c2aa70/include/gsl/gsl_util)

I've verified that this compiles with GCC and I've been using the resulting build fairly heavily the past few hours and I've not seen any issues (though with that said nothing is yet using Span directly either).

### DIFF
--- a/mfbt/Casting.h
+++ b/mfbt/Casting.h
@@ -238,6 +238,19 @@ AssertedCast(const From aFrom)
   return static_cast<To>(aFrom);
 }
 
+/**
+ * Cast a value of integral type |From| to a value of integral type |To|,
+ * release asserting that the cast will be a safe cast per C++ (that is, that
+ * |to| is in the range of values permitted for the type |From|).
+ */
+template<typename To, typename From>
+inline To
+ReleaseAssertedCast(const From aFrom)
+{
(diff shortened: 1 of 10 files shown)


wolfbeast:
I won't have time to look at this in-depth before my move. I'll just do a casual check and if it looks OK I'll merge it in so it can be used and you're not hung up on my delays.

As an aside: I find it weird that they feel the need to create a new data structure for this. Aren't the C++ standard constructs and `Mozilla::Range` enough? o.O

PR #826, opened 2018-10-08
Fix in-tab close button behavior

### LINKED ISSUE #775: Quickly pressing the close button on tabs has a delay which is not seen in Firefox or chrome

MaxKoll:
Fixes #775 (non-responsiveness when closing multiple tabs by rapidly clicking in-tab close buttons).

### DIFF
--- a/application/palemoon/base/content/tabbrowser.xml
+++ b/application/palemoon/base/content/tabbrowser.xml
@@ -4296,17 +4296,76 @@
             event.originalTarget.localName != "box")
           return;
 
-        // See hack note in the tabbrowser-close-tab-button binding
+        // See comments in the "mousedown" and "click" event handlers of the
+        // tabbrowser-tabs binding.
         if (!this._blockDblClick)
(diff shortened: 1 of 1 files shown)


mattatobin:
I hope this works on all three states of the tab close button and also works on both states for the tab bar and with and without animations.

Issue #816, opened 2018-10-07
ES6 stringification needs to be made spec compliant (ES2019 misc)


Sa-Ja-Di:
As Yami_ wrote: here:
https://forum.palemoon.org/viewtopic.php?f=3&t=20610#p154138

The News section issue was fixed in Firefox 55:
Last bad: 2017-04-17
First good: 2017-04-18
Pushlog
https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=9379831bb9c3d9abfea7dbf8dd06dbdab1d81dc4&tochange=bb38d935d699e0529f9e0bb35578d381026415c4

Perhaps that helps to limit the number of culprits.

Also, i have seen a warning in the error console which looks rather strange:

Timestamp: 07.10.2018 13:11:43
Warning: Unexpected end of file while searching for length value for matched media condition.
Source File: from DOM
Line: 0

Repeats a lot of times when visiting the Gog main page

kn-yami:
Actually the whole `iframe#js-promo-iframe` (including the _news_ section) is broken. I think I know why: one of the JS frameworks (Angular, to be precise) used on the web page displayed in this `iframe` fails to initialize. Why it fails to initialize? Because it tries to call `Function.prototype.apply.call(n, o, t)` with `n` being a class. This obviously fails, for example:
```
js> class C {};
js> Function.prototype.apply.call(C);
typein:2:1 TypeError: class constructors must be invoked with |new|
```
Why is tries to "execute" a class? Angular differentiates between classes and functions by looking for the `class` keyword in string representation of the given object. This fails in UXP version of SpiderMonkey, for example:
```
js> class C {};
js> C.toString();
"function C() {\n    \n}"
```
Why does string representation of a class is a function? Probably because _classes are in fact "special functions"_, [at least according to Mozilla](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#Defining_classes). The issue with string representation of classes was reported to Mozilla as [Bug 1216630](https://bugzilla.mozilla.org/show_bug.cgi?id=1216630) (commits that fixed the issue are in my pushlog), and to Angular as angular/angular.js#14240. SpiderMonkey versions with those fixes applied will behave like in this example:
```
js> class C {};
js> C.toString();
"class C {}"
```

wolfbeast:
Good analysis - thanks. Although I'm scratching my head as to why a framework would use toString() on a jsObj and then expect to use the result as a source for further processing...

[...]

PR #830, opened 2018-10-10
Several WebM/VPX fixes

### LINKED ISSUE #829: Add new helper functions to WebMDemuxer instead of calling libvpx directly.
In order to simplify adding additional decoders (e.g. AV1) in the future, we should add some helper functions to WebMDemuxer instead of calling libvpx directly. Additionally this will allow some logic to be simplified and improved when decoding WebM files.

Blocks #87

trav90:
This PR fixes/simplifies several things related to WebM and VPX playback:

- Correctly differentiate between a genuine EOS and an error when playing back WebM files.
- Simplify some logic when decoding WebM files.
- Add some new helper functions to WebMDemuxer instead of calling libvpx directly which will simplify adding other decoders (e.g. AV1) in the future.

Additionally, this PR adds code that begins using mozilla:Span that was added in #816.

I've tested this on Linux fairly extensively today and it all appears to be working as intended. Will wait for verification on Windows and will then push this out to unstable to cook a bit longer.

Resolves #829

### DIFF
--- a/dom/media/MediaData.h
+++ b/dom/media/MediaData.h
@@ -14,6 +14,7 @@
 #include "nsIMemoryReporter.h"
 #include "SharedBuffer.h"
 #include "mozilla/RefPtr.h"
+#include "mozilla/Span.h"
 #include "mozilla/UniquePtr.h"
 #include "mozilla/UniquePtrExtensions.h"
 #include "nsTArray.h"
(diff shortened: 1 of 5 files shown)


wolfbeast:
I'm merging this in and will build unstable (if it builds) but won't be able to test, so will just toss it to the sharks.

kn-yami:
> Additionally, this PR adds code that begins using mozilla:Span that was added in #816.

`mozilla:Span` was added in #818.

Issue #832, opened 2018-10-11
Basilisk: Remove in-browser translation


Ascrod:
I believe that this is dead code and can be removed. The feature was never completed in Firefox, the UI is hidden by default, and it relies on external translation services from Bing and Yandex to function (at least in theory -  in its present state, as far as I can tell, it just errors out).

I do not plan on removing the language detection module at this time, since that component is still used (albeit in a minor way) by the Reader View module in Basilisk.

I'm told that IceWeasel-UXP has already done this. Relevant commits for reference:
[765b196b](https://git.hyperbola.info:50100/software/iceweasel-uxp.git/commit/?id=765b196b2ae3d227d8e408cc44e0309fed77cd33)
[92d45a8f](https://git.hyperbola.info:50100/software/iceweasel-uxp.git/commit/?id=92d45a8fa9de61d89eeb03333f34f25e19de5fb6)
[eef869e9](https://git.hyperbola.info:50100/software/iceweasel-uxp.git/commit/?id=eef869e997ac5644ece7fc8ac9edaf51bd009931)
[d82ce5a7](https://git.hyperbola.info:50100/software/iceweasel-uxp.git/commit/?id=d82ce5a7142d5c45eee9d58c9784f02702a46716)

I can submit a PR for this.

wolfbeast:
Is this complete?

Ascrod:
Yes, sorry. I will close the issue.

JustOff:
String changes in this issue does not affect Pale Moon.
@wolfbeast, perhaps we don't need this tag here?

wolfbeast:
I'd like to keep the tag on anything that changes/adds/removes l10n strings, for potential other consumers of UXP.

JustOff:
Ok, thanks for the clarification.

PR #836, opened 2018-10-15
Basilisk: Remove in-browser translation

Ascrod:
This PR removes the half-baked in-browser translation feature. Code cleanup. Tag #832.


wolfbeast:
Asking or a review from JustOff (if you have time?) Since it's mostly fe stuff and I'm still rather unhanded without a workstation due to my move.

Issue #817, opened 2018-10-07
Non-working notifications


mimecry:
1. When I click on a notification, the focus is not switched to the tab of the notification's origin

2. The following button has no effect

![image](https://user-images.githubusercontent.com/10742312/46582350-646f5a80-ca6f-11e8-9d66-04e657bad104.png)

This error appears in the console when a notification appears. No other error message appears, despite my attempts to interact with the notification dialog.

![image](https://user-images.githubusercontent.com/10742312/46582367-a9938c80-ca6f-11e8-8cda-8cea2e73ed3c.png)

Windows 10 x64 1803, PM 28.1.0, fresh profile

Vangelis66:
This is a regression of Tycho issue [#1641](https://github.com/MoonchildProductions/Pale-Moon/issues/1641) due to change of platform to UXP; it is already reported in the forums at least twice: 

https://forum.palemoon.org/viewtopic.php?p=151120#p151120

> I think this fix was somehow lost in 28.0.x. After I updated, the pop-ups no longer switch me to the browser and/or tab.

and 

https://forum.palemoon.org/viewtopic.php?p=151769#p151769

> couple of notification bugs i noticed:
> - clicking on one doesn't direct you to the associated tab (TST bug?)
> - nothing happens when clicking on 'Notification settings'

but it went somehow unnoticed by the devs... FWIW, this is not a platform bug, but it is specific to application **Pale Moon**; Basilisk 52 behaves correctly... 

STR: 
OS: Win7 SP1 Home Premium 32-bit
Browser: Pale Moon 28.2.0a1 32-bit unstable, buildID=20181007063016
Test page: http://www.bennish.net/web-notifications.html

1. Click **Authorize** and grant notifications permissions (either for session/permanently)
2. Click **Show**; a web notification is shown for some seconds in the lower right of the GUI; if you are quick enough to click the gear button => Notification settings, this does not bring up any notification settings dialogue, but instead hides the whole notification...
(NB: On **Basilisk**, clicking the gear button takes you to **about:preferences#content** (in-content prefs tab), where you can indeed access **Notification Settings**;
Pale Moon does not have in-content prefs (Options), nor, to my knowledge, any equivalent notification setting in its GUI...
[...]

PR #141, opened 2018-04-13
 Fix: Findbar broken

janekptacijarabaci:
This resolves #138 (partially)

---

__TODO (I leave for you to correct - I don't know what your preferences are...):__
https://github.com/MoonchildProductions/UXP/pull/141/files#diff-35cdf6e12e681e90421415d441835592R3801

`accessibility.typeaheadfind.highlightallremember`
and
`accessibility.typeaheadfind.highlightallbydefault`
does not exist (now)

Ad https://github.com/MoonchildProductions/Pale-Moon/issues/364
An actual preference: `findbar.highlightAll`

---

Bug(s):
https://bugzilla.mozilla.org/show_bug.cgi?id=1133981
https://bugzilla.mozilla.org/show_bug.cgi?id=1154799
https://bugzilla.mozilla.org/show_bug.cgi?id=1320095

Also:
http://xref.palemoon.org/uxp-trunk/source/toolkit/modules/Finder.jsm#49

---

I've created the new build (x32, Windows) - `Pale Moon UXP` - and tested.

PR #838, opened 2018-10-20
Pale Moon: Remove message manager property from tabbrowser.xml

FranklinDM:
This removes the `messageManager` property from Pale Moon's `tabbrowser.xml`.

It also seems to solve the first point of #817 on my side (focus is not switched to the tab of the notification's origin).

Please add `Verification Needed` tag as I've not tested this on any other platform.

### DIFF
--- a/application/palemoon/base/content/tabbrowser.xml
+++ b/application/palemoon/base/content/tabbrowser.xml
@@ -2950,19 +2950,6 @@
                 onget="return this.mCurrentBrowser.docShell"
                 readonly="true"/>
 
-      <property name="messageManager"
-                readonly="true">
-        <getter>
-          <![CDATA[
-            let frameLoader = this.mCurrentBrowser.frameLoader;
-            if (!frameLoader) {
-              return null;
-            }
-            return frameLoader.messageManager;
-          ]]>
-        </getter>
-      </property>
-
       <property name="webNavigation"
                 onget="return this.mCurrentBrowser.webNavigation"
                 readonly="true"/>


JustOff:
Are you really sure this is a good idea? I suspect it may break many things, while the positive effect on #817 is simply coincidental.

FranklinDM:
I researched even further on how this code got into our tree (I should've included this in the main PR):
- This wasn't present neither in Tycho nor in any Firefox code (checked the history of tabbrowser.xml) since `messageManager` started to be used in tabbrowser.xml. 
  - Final PM-Tycho commit involving this file: MoonchildProductions/Pale-Moon@59a8c7bfb05e6151b9b768e7af69f810df28cbd6.
  - First time `messageManager` was used: [bug 662008](https://bugzilla.mozilla.org/show_bug.cgi?id=662008).
- This got into Pale Moon 28 through 2e0e50636fb38354ca4957e3713dd8434e0347ed (findbar code related).
- Based on the message at PR #141, [bug 1154799](https://bugzilla.mozilla.org/show_bug.cgi?id=1154799) that doesn't apply to us (because it's for Thunderbird, chat/IM component specifically) might have been included in the aforementioned commit. Our browser binding implements `nsIMessageListener` unlike Thunderbird.
- The effect wasn't much visible because:
  - AFAIK, it only has 2 consumers.
  - DOMTitleChanged only gets raised when `browser.tabs.remote` is enabled. This pref is disabled by default in Pale Moon.
  - DOMWebNotification gets raised when clicking notifications/alerts.
- For reference, Basilisk's `tabbrowser.xml` doesn't have that code.

JustOff:
@FranklinDM, yes, I did the same research in parallel and it seems my assertion above was wrong. The only thing I still don't understand is why this code was added in #141.

wolfbeast:
Why it was added was mode code-rolling by @janekptacijarabaci who has since disappeared. so... if it can go, then by all means, I'm all for it.

PR #839, opened 2018-10-20
remove named function syntax in several TychoAM components

kn-yami:
I did this more than two weeks ago and then forgot about it...
Meta issue: #516.

PR #843, opened 2018-10-20
fix naming of getStartupChanges function

kn-yami:
Fixes the error reported by @JustOff at #839. I still have no idea how did I miss this obvious issue. Sorry.


JustOff:
I confirm this resolves the issue.

PR #847, opened 2018-10-23
Don't use NS_ENSURE_SUCCESS where it causes useless logspam.

g4jc:
Log spam fix. Suppresses 1,100 instances of NS_ENSURE_SUCCESS(rv, rv) failed with result 0x80520012 emitted from netwerk/base/nsIOService.cpp during linux64 debug testing. Backports [m-c 1333174](https://bugzilla.mozilla.org/show_bug.cgi?id=1333174).


wolfbeast:
While you're doing this, could you add the same fix for `nsNetUtilInlines.h`? For some reason these are separate bugs at Mozilla but should be the same (they perfectly correlate; exact same issue, exact same patch, just different location).
[Bug 1333172](https://bugzilla.mozilla.org/show_bug.cgi?id=1333172)

g4jc:
Sure. Done. 👍

PR #850, opened 2018-10-24
Add initial support for AV1 video.

### LINKED ISSUE #87: Support AV1 video

trav90:
This PR adds libaom and provides initial support for AV1 video playback. With one exception (see below), all code introduced in this PR is conditional and is not built by default (--enable-av1 must be specified in mozconfig). Additionally, even when AV1 support is built it is behind a pref ("media.av1.enabled") which is disabled by default.

The single non-conditional change made by this PR is increasing the stack size for media decoders to prevent stack overflows. This is a low-risk change that I have tested fairly extensively on Linux (playing back videos from YouTube, Twitch etc.).

While several test AV1-encoded WebM videos play fine with these patches, AV1 in MP4 and AV1 with MSE is not supported yet (I will create a followup issue for this). It should be noted that because of this the AV1-encoded playlist on YouTube does not work yet.

Also worth noting, Mozilla reported OOM errors when building libaom for 32-bit Windows using MSVC. The report they gave was very vague and only seemed to be reproducible by one person, but I can't verify this without a Windows environment.

Below are several links to AV1-encoded WebM videos used by both Mozilla and Google for testing during their respective implementations:

- https://raw.githubusercontent.com/chromium/chromium/b430c9d0c83a00fef1ecdefd8c0d83f935c68b24/media/test/data/bear-av1-480x360.webm
- https://raw.githubusercontent.com/chromium/chromium/b430c9d0c83a00fef1ecdefd8c0d83f935c68b24/media/test/data/bear-av1-640x480.webm
- https://raw.githubusercontent.com/chromium/chromium/b430c9d0c83a00fef1ecdefd8c0d83f935c68b24/media/test/data/bear-av1.webm
- https://demo.bitmovin.com/public/firefox/av1/

Resolves #87 (with new issue to follow for AV1 in MP4 + MSE).


mattatobin:
Jesus. Anyway, OOM for Win32 (if it is ran into) could be avoided by getting `gkmedias` split back out from `libevil`.

wolfbeast:
I'll build a new unstable today with these changes in to toss to the sharks.

PR #862, opened 2018-11-03
[PALEMOON] Focus and select the first non-collapsed text element in the bookmark properties dialog

### LINKED ISSUE #861: The bookmark properties window should focus on the Name Picker text box on init
Steps:
* The Bookmarks sidebar is open
* I do "right-click > New Folder..."
* There is no focus, I have to click in the input before typing the name

Adding this focus would make the process much more convenient!

Other notes:
* For consistency, the same should be done with the entry "right-click > New Bookmark..."
* Firefox already has these focus

<br> Pale Moon 28.1.0 - Windows 7

JustOff:
This resolves #861.

As a positive side effect, this will also focus the search field when the library window opens.

### DIFF
--- a/application/palemoon/components/places/content/editBookmarkOverlay.js
+++ b/application/palemoon/components/places/content/editBookmarkOverlay.js
@@ -222,6 +222,11 @@ var gEditItemOverlay = {
     }
 
     let focusElement = () => {
+      let elt = document.querySelector("textbox:not([collapsed=true])");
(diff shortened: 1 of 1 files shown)


wolfbeast:
the "positive side effect" is a negative side effect, since the focus remains forced on it.

JustOff:
You are right, I overlooked this, sorry.

PR #866, opened 2018-11-08
Make older autocomplete default and new autocomplete australis-only

mattatobin:
I'd appreciate a cursory review on ` toolkit/components/satchel/AutoCompletePopup.jsm`


mattatobin:
I am shelving this for now..

PR #876, opened 2018-11-18
Add support for AV1 in MP4 container

### LINKED ISSUE #853: Add support for AV1 in MP4 + MSE
Now that #87 is resolved and we have initial support for AV1 in WebM, we should also add support for AV1 in MP4 + MSE. This will allow AV1 video playback on sites such as YouTube.

trav90:
This PR updates our in-tree copy of libstagefright (used by UXP applications for MP4 parsing) to support AV1 video within the MP4 container. It also adds AV1 support to the MP4Decoder (allowing playback of AV1-encoded MP4 files). This PR resolves #853 and allows for AV1 in MP4 + MSE playback. 480p playback seemed to work just fine, however I did notice some performance issues when playing back 720p and 1080p video on YouTube. I'm thinking that it is due to the higher performance demands of AV1 decoding using libaom, but will investigate (I'd be interested to hear if others also experience performance issues).

Some links for testing AV1 in MP4:

[Test file from Google.](https://raw.githubusercontent.com/chromium/chromium/3033eaecc684727a6f87834d3336a509f9ec9e36/media/test/data/bear-av1.mp4)
[Test file from Mozilla.](https://hg.mozilla.org/mozilla-central/raw-file/efb8bd6f8062/dom/media/test/av1.mp4)
[Link to enable AV1 playback on YouTube.](https://www.youtube.com/testtube)
[Link to AV1 playlist on YouTube.](https://www.youtube.com/watch?v=2nXYbGmF3_Q&t=0s&list=PLyqf6gJt7KuHBmeVzZteZUlNUQAVLwrZS&index=2)

### DIFF
--- a/dom/media/fmp4/MP4Decoder.cpp
+++ b/dom/media/fmp4/MP4Decoder.cpp
@@ -139,6 +139,14 @@ MP4Decoder::CanHandleMediaType(const MediaContentType& aType,
             NS_LITERAL_CSTRING("audio/flac"), aType));
         continue;
       }
+#ifdef MOZ_AV1
+      if (IsAV1CodecString(codec)) {
+        trackInfos.AppendElement(
+          CreateTrackInfoWithMIMETypeAndContentTypeExtraParameters(
+            NS_LITERAL_CSTRING("video/av1"), aType));
+        continue;
+      }
+#endif
       // Note: Only accept H.264 in a video content type, not in an audio
       // content type.
       if (IsWhitelistedH264Codec(codec) && isMP4Video) {
--- a/media/libstagefright/frameworks/av/include/media/stagefright/MediaDefs.h
+++ b/media/libstagefright/frameworks/av/include/media/stagefright/MediaDefs.h
@@ -25,6 +25,7 @@ extern const char *MEDIA_MIMETYPE_IMAGE_JPEG;
 extern const char *MEDIA_MIMETYPE_VIDEO_VP6;
 extern const char *MEDIA_MIMETYPE_VIDEO_VP8;
 extern const char *MEDIA_MIMETYPE_VIDEO_VP9;
+extern const char *MEDIA_MIMETYPE_VIDEO_AV1;
 extern const char *MEDIA_MIMETYPE_VIDEO_AVC;
 extern const char *MEDIA_MIMETYPE_VIDEO_MPEG4;
 extern const char *MEDIA_MIMETYPE_VIDEO_H263;
--- a/media/libstagefright/frameworks/av/media/libstagefright/MPEG4Extractor.cpp
+++ b/media/libstagefright/frameworks/av/media/libstagefright/MPEG4Extractor.cpp
@@ -269,6 +269,9 @@ static const char *FourCC2MIME(uint32_t fourcc) {
         case FOURCC('V', 'P', '6', 'F'):
             return MEDIA_MIMETYPE_VIDEO_VP6;
 
+        case FOURCC('a', 'v', '0', '1'):
+            return MEDIA_MIMETYPE_VIDEO_AV1;
+
         default:
             ALOGE("Unknown MIME type %08x", fourcc);
             return NULL;
--- a/media/libstagefright/frameworks/av/media/libstagefright/MPEG4Extractor.cpp
+++ b/media/libstagefright/frameworks/av/media/libstagefright/MPEG4Extractor.cpp
@@ -270,6 +270,7 @@ static const char *FourCC2MIME(uint32_t fourcc) {
             return MEDIA_MIMETYPE_VIDEO_VP6;
 
         case FOURCC('a', 'v', '0', '1'):
+        case FOURCC('.', 'a', 'v', '1'):
             return MEDIA_MIMETYPE_VIDEO_AV1;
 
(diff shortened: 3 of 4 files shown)


JustOff:
Unfortunately, I see a lot of freezes in different resolutions and bitrates on both [YouTube](https://www.youtube.com/watch?v=2nXYbGmF3_Q&t=0s&list=PLyqf6gJt7KuHBmeVzZteZUlNUQAVLwrZS&index=2) and [Bitmovin](https://demo.bitmovin.com/public/firefox/av1/), while the processor load does not exceed 50-60%.

trav90:
Does audio and video freeze or just audio?

On the Linux box I tested on video would sometimes freeze while the audio would continue playing normally. CPU was around 80-90% across all cores.

trav90:
I just tested a few other videos on YouTube at 720p and (intermittently) had the same issue. 480p worked perfectly.

I can't reproduce the freezing at Bitmovin (though these patches should have no affect as the Bitmovin demo plays a WebM video and not MP4).

I really think that it the video freezing is simply due to the demands of AV1 decoding using libaom (which should improve with time by upstream). It is something I will continue to investigate, but I don't think it should block this being merged (our AV1 playback is considered to be experimental at this point).

trav90:
FWIW I just tested the YouTube playlist using the latest Firefox Nightly build and encountered very similar video freezing and high CPU usage as with the current UXP trunk + these patches.

wolfbeast:
I'm sorry, but... what exactly is the point of a new codec if decoding with it is so expensive that you can't even play 720p on modern hardware with it without choking?

JustOff:
It seems in my case the problem is 32-bit build of Pale Moon, as in 64-bit one Bitmovin plays just perfect. I did not build 64-bit Pale Moon with this PR yet, but I guess the results with YouTube will be just the same as I get with Firefox Nightly 64-bit: all demos up to 720p plays fine, 1080p are choppy.

I also found that in Firefox Nightly 32-bit AV1 playback is hard disabled, perhaps it makes sense for us to do the same.

trav90:
@wolfbeast that is a good question that nobody else seems to be asking on the web. I’ve found very few articles anywhere that mentions the high cost of decoding AV1.


@JustOff Mozilla disables AV1 on 32-bit Windows due to linking issues (rather than attempt to fix it they just disabled it on that platform).

JustOff:
Well, so now we also know that 32-bit builds have additional (performance?) issues that Mozilla probably isn't aware of yet.

wolfbeast:
@Nuck-TH http://www.palemoon.org/unstable/ -- have fun with it!

Sa-Ja-Di:
Tested the new coded with the 64 Bit and the Youtube playlist - so far i have seen no freezes.

trav90:
@Sa-Ja-Di AV1 decoding should work both with and without MSE.

kn-yami:
It works without MSE - it is YouTube's decision to only serve it when your browser supports MSE. Also for me it drops frames even in `720p60`.

NintendoManiac64:
I realize I'm a bit late to this discussion, but there's been a lot of discussion around AV1 decoder performance over on the Doom9 forums.

The jist is that libaom is first and foremost a reference decoder that is very much not optimized for performance (but rather presumably code readability?).  Already, in terms of decoding performance, the new dav1d AV1 decoder is leaps and bounds faster with much better multithreading support as well (libaom can only really make good use of 2 CPU threads).

trav90:
It is true that libaom is first and foremost a reference decoder; though with that said the current version we use does provide better performance than previous versions so improvements on that front are still happening upstream.

Looks like Mozilla is looking at dav1d as well so I will keep an eye on it. I don’t particularly want to import and maintain two different libs for decoding the same codec, but if dav1d does seem to perform better on my systems (one of which does not have an AVX/AVX2 capable CPU), then I will evaluate using it instead of libaom.

kn-yami:
Well, that is what I suspected.

NintendoManiac64:
It is my understanding that AVX optimizations were the primary focus up until now, and that SSE and ARM optimizations are to follow.

(for reference, I don't even own a CPU that supports AVX)

trav90:
@NintendoManiac64 yes that is correct. For what it’s worth SSE3 assembly optimizations for dav1d landed earlier today so I’m assuming SSE2 will be next.

Another thing to note: dav1d requires nasm for it’s assembly optimized code and is apparently not buildable with yasm which is what UXP currently uses. We would likely need to follow suite with Mozilla in switching to nasm with yasm being used as a fallback.

wolfbeast:
I'm not comfortable swapping out yasm just so an alternative decoder lib can be built - many critical paths in our tree (especially in graphics rendering, filtering, etc.) use assembly and switching compiler is like touching every line of code. It will likely also impact _other_ codecs that use assembly like VP8/VP9 (ffmpeg/libvpx). Mozilla has in recent patches also shown that their work is largely becoming incompatible with our code and causing additional problems, so another caution there.

wolfbeast:
Let's continue this discussion in #888 instead of on a closed/merged PR

mattatobin:
Is there any reason both assembly compilers can't be used or that the assembly can't be made yasm compatibile?

Issue #288, opened 2018-04-28
Remove GONK widget code


wolfbeast:
Since B2G has been nuked, the widget code for it can go.

- [x] Remove all conditional `MOZ_WIDGET_GONK` ifdefs and code blocks
- [x] Remove `/widget/gonk/*`
- [x] Remove `MOZ_B2G*`
- [x] Convert uses of `KEY_MAP_ANDROID_EXCEPT_B2G` to `KEY_MAP_ANDROID`
- [x] Remove other condition gonk widget code and unused files
- [x] Remove `moz_widget_toolkit_gonk`

wolfbeast:
Total score: 139695+5017+6+13+750+1357 = 146838 axed lines of dead code.

JustOff:
It seems you missed `MOZ_WIDGET_GONK` in `Performance.cpp` [here](https://github.com/MoonchildProductions/UXP/blob/e16bcd08aae85a7d9c2de5a4b1c733280cb81112/dom/performance/Performance.cpp#L30).

Also, I'm not sure about [MOZ_WIDGET_TOOLKIT == gonk](http://xref.palemoon.org/uxp-trunk/search?string=%28%22gonk%22|%27gonk%27%29&regexp=1&find=&findi=&filter=^[^\0]*%24&hitlimit=&tree=uxp-trunk) and [MOZ_GONK_MEDIACODEC](http://xref.palemoon.org/uxp-trunk/search?string=MOZ_GONK_MEDIACODEC&find=&findi=&filter=^[^\0]*%24&hitlimit=&tree=uxp-trunk).

wolfbeast:
Strange -- I know for a fact I fixed that.
As for the other two, the first I didn't think about checking. the second isn't widget code but should also be cleaned up (separate issue)

JustOff:
I actually have a half of [MOZ_WIDGET_TOOLKIT == gonk](http://xref.palemoon.org/uxp-trunk/search?string=%28%22gonk%22%7C%27gonk%27%29&regexp=1&find=&findi=&filter=%5E%5B%5E%5C0%5D*%24&hitlimit=&tree=uxp-trunk) fixes already done in my test branch, so I can finish it and submit as a PR if you wish.

[...]

PR #879, opened 2018-11-21
Remove leftover gonk bits in /dom/media

trav90:
Just what it says in the tin.

Tag #288


dom/media/platforms/PDMFactory.cpp
```diff
@@ -19,9 +19,6 @@
 #ifdef MOZ_APPLEMEDIA
 #include "AppleDecoderModule.h"
 #endif
-#ifdef MOZ_GONK_MEDIACODEC
-#include "GonkDecoderModule.h"
```
wolfbeast: Did you also remove this header file which is obviously no longer needed?

trav90: That header file was already removed previously as part of the main gonk cleanup.

Issue #21, opened 2018-02-18
Remove Telemetry


wolfbeast:
This is going to be a big task, to be taken in small steps to not bust everything.

JustOff:
I accidentally found that health report service is enabled for Pale Moon, because `MOZ_SERVICES_HEALTHREPORT=1` by default in `old-configure.in`.  I'm not sure if any data are actually send to Mozilla, but it is at least recording in `saved-telemetry-pings` folder in the profile.

wolfbeast:
Since there is no release of any kind of Pale Moon on UXP yet, what is your issue?

mattatobin:
Part of the first step to porting was to match configuration where possible to eliminate issues.. We know Mozilla has botched conditional status of platform components so working out those bits while trying to achieve basic buildablity work against each other.

Rest assured, this crap won't stick around for long.

wolfbeast:
I've pushed a few conf changes in confvars so your precious data won't be collected in a health report by default.

JustOff:
Thanks, this is exactly what I wanted to have until the telemetry is wiped off completely.

THEtomaso:
I understand that this will involve a lot of work.
..but shouldn't this issue be tagged for the 28.0.0 release milestone?
Many users will probably want to stay with Tycho, until it has been resolved.

wolfbeast:
It's not a release blocker. Telemetry is disabled -- actually removing it from the tree is something that will take time but is ultimately just code cleanup.

JustOff:
@THEtomaso, JFYI: Tycho also has Telemetry disabled, and not removed.

THEtomaso:
Thanks for clarifying, guys. :)

THEtomaso:
Another question though..
If it's just a matter of code cleanup, why is this issue still labeled with "High Risk", "Privacy" and "Security"?

wolfbeast:
The code removal is high risk because it is weaved into many things and can break functionality or impact security if not carefully removed. And it's telemetry, to labeled with Privacy.

wolfbeast:
Note to self: remove PaintTelemetry from /layout.

Kokokokoka:
There was #1074 which was closed. I would like to know, how could one fix this PR in order for it to be accepted?

adeshkp:
@Kokokokoka By not doing everything in one single PR.
Do it step by step if interested, otherwise I'll take care of it eventually (too busy at the moment).

[...]

PR #897, opened 2018-12-02
Remove MediaTelemetryConstants

trav90:
MediaTelemetryConstants is Windows-only telemetry used by Mozilla to determine the version of WMF used for decoding. This is unused code in our tree (the code that actually used this was removed previously), so this PR simply removes the file and includes. Still adding the `Verification Needed` label just in case since I do not have a Windows build environment.

Tag #21


JustOff:
I haven't tried to make clobber, but otherwise Pale Moon is built w/o any problems on Windows.

PR #898, opened 2018-12-02
Remove FxA Phase 1

wolfbeast:
Considering the number of changes, making this a pull request -- preferably build on all supported platforms and test against the Pale Moon Sync server from Pale Moon and Basilisk.

- Reverted the sync client to Tycho's version
- Removed several FxA integrations from the sync client
- Changed Basilisk FE code to work with Sync 1.1/Weave, which was severely crippled or simply not implemented properly, but panel code was still left in which could thankfully be re-used.

I will have to take a break from this code after phase 1 because I can't bear to look at the CUI and FxA code anymore at this moment after needing a lot of trial&error and rebuilds. The remaining code is still very much entangled because of the way FxA was implemented in Firefox with little regard to proper structure or coding practices. Essential parts of the interface between FE and sync client run through FxA modules or script files that should be pulled apart and reorganized if this is to be cleaned up properly in phase 2.

Tagging JustOff, Tobin and Travis for review. if you have time, please look over the code changes. If you don't that's fine and simply approve with a cursory check.


trav90:
Sorry for the delay in reviewing this. I'm going to try to get to it today.

wolfbeast:
Thanks for reviewing/testing, Travis.
Hmm.. Not sure why View Quota would not be there, I'll have a look.
Also, syncing add-ons should be removed from Basilisk. that has never worked right and requires an add-on storage server that's no longer available anyway.

wolfbeast:
Looks like "Quota" was just removed -- I guess because FxA/Sync 1.5 doesn't have a quota system in the weave way so why not break half the functionality XD. I'll slot that back in.

wolfbeast:
I'll just be merging this in then.

JustOff:
String changes were pushed to CrowdIn.

PR #914, opened 2018-12-18
Fix false positives in the preprocessor-checker

### LINKED ISSUE #912: Avoid false positives in preprocessor-checker
At the end of our build process, we have a check that verifies if all files in the objdir that need preprocessing are actually preprocessed.
In Interlink, this causes a false positive error because a `%`-prefixed DTD reference in an XML file is seen as a preprocessor command.

`%`-prefixed commands are only valid in CSS -- checks for that should be restricted to CSS. Similarly, `#`-prefixed commands should be restricted to other files and not checked for in CSS files. I.e.: the script needs to be a little smarter.

@JustOff this is your script; assigning you to improve it :)

Tag #429.

JustOff:
This resolves #912.


wolfbeast:
I'll just have to trust this patch works -- this is way beyond me.

PR #916, opened 2018-12-22
undepercate search service synchronous init

kn-yami:
details: https://forum.palemoon.org/viewtopic.php?p=158385#p158387

### DIFF
--- a/toolkit/components/search/orginal/nsSearchService.js
+++ b/toolkit/components/search/orginal/nsSearchService.js
@@ -19,8 +19,6 @@ XPCOMUtils.defineLazyModuleGetter(this, "OS",
   "resource://gre/modules/osfile.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "Task",
   "resource://gre/modules/Task.jsm");
-XPCOMUtils.defineLazyModuleGetter(this, "Deprecated",
-  "resource://gre/modules/Deprecated.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "SearchStaticData",
   "resource://gre/modules/SearchStaticData.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "setTimeout",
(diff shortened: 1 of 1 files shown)


wolfbeast:
No, sorry. Not taking this.
Asynchronous init on startup has been a thing since Firefox 16. People should use the proper method when calling the service on startup.

kn-yami:
OK.

JustOff:
It looks a bit inconsistent, considering https://github.com/MoonchildProductions/UXP/commit/a930a790d53f82c22167eec57d24eec943298968 (this is currently [used in Basilisk/IceWeasel/IceDove](https://github.com/MoonchildProductions/UXP/commit/9fe00e826d2a5f57d47a598c6833df00a60fe3cb#diff-6cfe79b8cfbc3ac887d6974b9cb4f0dd)). I think it would be logical to deprecate this or not the same for all UXP applications.

wolfbeast:
I've discussed this with Tobin over jabber and I think a warning should be given in this particular case, because the search service _is_ heavy and initializing it this way will pull in a lot of code in a blocking way, but should be stated differently (as a performance warning, not a deprecation warning-error). Once done that should apply to all UXP applications.

wolfbeast:
@kn-yami Sorry for getting our wires crossed here, I didn't see the forum discussion before seeing this PR out of the blue. 
You're right and this should simply be removed, but we do need to find a way to report performance issues to the user instead of just allowing things like this silently. That can be done as a follow-up issue, I guess.

JustOff:
> the search service is heavy and initializing it this way will pull in a lot of code in a blocking way

If this is the real thing, then we probably also need a separate issue to change the way the quickdial page initializes, because currently this deprecation warning is triggered when the quickdial is set as the default for new tabs.

mattatobin:
I think the main problem with the synchronous init shows its self the most if say the browser plus new tab page plus two or three extensions were to all call and init it for their selves in rapid succession.

One rouge invokation isn't likely to make much of a noticeable impact as a cumulative impact by several things all doing it at startup.

wolfbeast:
> If this is the real thing, then we probably also need a separate issue to change the way the quickdial page initializes, because currently this deprecation warning is triggered when the quickdial is configured to display by default on startup.

Opened #918 for that, that needs to be fixed.

wolfbeast:
> Deprecated.jsm could be adapted to show performance warnings

Instead of a separate module, we should probably just make a separate function inside Deprecated - these performance warnings are likely going to always be because something is deprecated... right? Or is it likely we want to issue performance warnings this way for things other than deprecated functions?

mattatobin:
I figure just do it in `Deprecated.jsm` for now.. It's usage and a performance warning are low usage as it is.. Mostly in places and a few other modules.

I can take this if you want.

wolfbeast:
I've already poked at it, I'll do it myself -- shouldn't take much time anyway.

Issue #915, opened 2018-12-18
Use the new FFmpeg decode API with recent FFmpeg version


trav90:
Beginning with libavcodec 58 (which we now use after #806), the older avcodec_decode_video2 API is broken and will in some instances cause frames to be dropped (giving the appearance of stuttering) when playing VP9 video.

Our FFmpeg code needs to be updated so that libavcodec 58+ will use the new API.

wolfbeast:
Assuming this is solved by way of #921

PR #921, opened 2018-12-29
Use new decode API with recent FFmpeg version

trav90:
This PR updates our FFmpeg code to use the new decoding API used by libavcodec 58. This fixes the terminal warnings about the previous API being deprecated (as reported on the [forum](https://forum.palemoon.org/viewtopic.php?f=57&t=21016&p=158181#p157436)), however verification is needed on whether this fixes the VP9 stuttering in #915 (I personally can't see any discernible differences in VP9 playback between Pale Moon 28.1, 28.2, and a trunk build with these patches).


wolfbeast:
I'll toss out some unstables with these patches in, and the people who reported the stuttering can verify it.

wolfbeast:
Unstables are up.

trav90:
Linux unstables are up.

Issue #930, opened 2019-01-16
[Pale Moon] Option for disabling HSTS sets the wrong preference


Ascrod:
While poking through the HSTS code, I noticed that the checkbox in Pale Moon's security preferences for enabling or disabling the use of HSTS appears to toggle the use of the preload list, not the entire HSTS feature as the box's label seems to indicate. Personally I'm all for using HSTS all the time, but if this option is going to be available, I think it should either do what it says or say what it does.

To make this checkbox disable the entire HSTS feature, I believe [this commit](https://github.com/MoonchildProductions/Pale-Moon/commit/9bc65e235b62c4e84c69f301bd89de29769f4abf) can be ported from Tycho, and the checkbox's preference changed from 'preloadlist' to 'enabled'.

Thoughts? I can submit a PR for this if desired.

wolfbeast:
Hmm.. not sure where this went wrong -- it should enable/disable the entire feature, not the preloadlist.

Feel free to offer a PR.

wolfbeast:
Seems this was a mistake made by @trav90 in e370a636baa5435c7c57a3aebf2fca647023b7c8 -- assuming the pref was renamed while it was actually a new pref introduced in Tycho.
Thanks again for catching this -- I certainly wouldn't have ever noticed.

trav90:
Yeah, that was my mistake.. Thanks for catching @Ascrod.

Ascrod:
Happy to help @trav90. This should be fixed with PR #931, so I'll close this issue.

PR #931, opened 2019-01-17
Fix option for disabling HSTS in Pale Moon

Ascrod:
This addresses #930.

In addition to porting [this commit](https://github.com/MoonchildProductions/Pale-Moon/commit/9bc65e235b62c4e84c69f301bd89de29769f4abf), I made three additional changes to nsSiteSecurityService:
- Added a check in `SetHSTSState` so that the cache is not actually written to.
- Added a condition to the check in `IsSecureHost` so as not to interfere with HPKP.
- Added the preference to `Observe` so that it can take effect without restarting the browser.

### DIFF
--- a/application/palemoon/components/preferences/security.xul
+++ b/application/palemoon/components/preferences/security.xul
@@ -43,8 +43,8 @@
 
       <!-- Security Protocols -->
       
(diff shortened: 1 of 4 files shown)


Ascrod:
Yes, sorry about that. Fixed.

PR #945, opened 2019-01-26
Only use white audio indicator when using a lightweight theme

### LINKED ISSUE #942: incorrect display of the sound icon on the tab
#638 This is probably a bug, due to the Windows 10 theme.
She white. And it's hard to notice. Theme standard (for win 10), version of 28.3.1
default

When switching to another window, the icon is displayed correctly.
default
 @JustOff  @wolfbeast @mattatobin

FranklinDM:
Fixes #942.

The audio indicator's color should only be inverted when using a lightweight theme since it's the only time where tabs become semi-transparent.


bogachenko:
Updated omni.ja and everything seems to work correctly.

PR #959, opened 2019-02-02
Use 'mousemove' events for hiding the Navigation Toolbar in Full Screen

### LINKED ISSUE #336: Full-screen nav bar doesn't hide when focused during switching.
Separated from [#270](https://github.com/MoonchildProductions/UXP/issues/270#issuecomment-385694204), since it's not a release blocker:

> The CSS animation [..] isn't being triggered at all for blank (or about:logo) pages. When that happens even if you switch to a tab with content it doesn't ever trigger the hiding..

This issue exists in both Basilisk and Pale Moon (and Firefox Nightly 20180506).

FranklinDM:
Resolves #336.

* This involves using `mousemove` events instead of `MousePosTracker` for hiding the navigation toolbar in full screen mode.
* Similar to the previous implementation.
* [Pale Moon] Side-effect: Fixes double-F6 press to show location bar while in full screen.

Please add the `Verification Needed` tag.

### DIFF
--- a/application/basilisk/base/content/browser-fullScreenAndPointerLock.js
+++ b/application/basilisk/base/content/browser-fullScreenAndPointerLock.js
@@ -322,15 +322,9 @@ var FullScreen = {
       document.addEventListener("popuphidden", this._setPopupOpen, false);
       // If it is not safe to collapse, add the mouse position tracker or
       // else it won't be possible to hide the navigation toolbox again
-      if (!this._safeToCollapse()) {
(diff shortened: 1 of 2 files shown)


JustOff:
Sorry, but I don't like this approach, see https://github.com/MoonchildProductions/UXP/issues/336#issuecomment-459957352.

wolfbeast:
Looks like this is the best approach so far. Merging.

PR #967, opened 2019-02-09
Pale Moon: Add loop context menu to HTML5 media elements.

### LINKED ISSUE #955: Add Loop context menu item to html5 player
It would be nice to have.

Ascrod:
What it says on the tin. Resolves #955.

### DIFF
--- a/application/palemoon/locales/en-US/chrome/browser/browser.dtd
+++ b/application/palemoon/locales/en-US/chrome/browser/browser.dtd
@@ -433,6 +433,8 @@ fast. -->
 <!-- LOCALIZATION NOTE: The access keys for "Show Controls" and
 "Hide Controls" are the same because the two context-menu
 items are mutually exclusive. -->
+<!ENTITY mediaLoop.label             "Loop">
+<!ENTITY mediaLoop.accesskey         "L">
 <!ENTITY mediaShowControls.label     "Show Controls">
 <!ENTITY mediaShowControls.accesskey "C">
 <!ENTITY mediaHideControls.label     "Hide Controls">
--- a/application/palemoon/locales/en-US/chrome/browser/browser.dtd
+++ b/application/palemoon/locales/en-US/chrome/browser/browser.dtd
@@ -417,8 +417,8 @@ items are mutually exclusive. -->
 <!ENTITY mediaMute.accesskey         "M">
 <!ENTITY mediaUnmute.label           "Unmute">
 <!ENTITY mediaUnmute.accesskey       "m">
-<!ENTITY mediaPlaybackRate.label     "Play Speed">
-<!ENTITY mediaPlaybackRate.accesskey "l">
+<!ENTITY mediaPlaybackRate2.label     "Play Speed">
+<!ENTITY mediaPlaybackRate2.accesskey "d">
 <!ENTITY mediaPlaybackRate050x.label "Slow Motion (0.5×)">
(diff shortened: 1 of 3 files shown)


application/palemoon/locales/en-US/chrome/browser/browser.dtd
```diff
@@ -433,6 +433,8 @@ fast. -->
 <!-- LOCALIZATION NOTE: The access keys for "Show Controls" and
 "Hide Controls" are the same because the two context-menu
 items are mutually exclusive. -->
+<!ENTITY mediaLoop.label             "Loop">
+<!ENTITY mediaLoop.accesskey         "L">
```
JustOff: Please move these strings above LOCALIZATION NOTE related to "Show Controls" and "Hide Controls"

Ascrod: Whoops, missed that part. Should be fixed now.

mattatobin:
Basilisk?

application/palemoon/locales/en-US/chrome/browser/browser.dtd
JustOff: Thanks.

application/palemoon/locales/en-US/chrome/browser/browser.dtd
```diff
@@ -430,6 +430,8 @@ movie "Space Balls" and is meant to say that this speed is very
 fast. -->
 <!ENTITY mediaPlaybackRate200x.label "Ludicrous Speed (2×)">
 <!ENTITY mediaPlaybackRate200x.accesskey "L">
+<!ENTITY mediaLoop.label             "Loop">
+<!ENTITY mediaLoop.accesskey         "L">
 <!-- LOCALIZATION NOTE: The access keys for "Show Controls" and
```
JustOff: Isn't there a conflict between `mediaLoop.accesskey` and `mediaPlaybackRate.accesskey`?

Ascrod: Technically multiple items are allowed to have the same access key, you would just have to hit the key multiple times. But in this case the playback speed items are in a separate submenu, so that shouldn't be a problem here.

Ascrod:
@mattatobin Basilisk already has this feature. I checked. :wink:

application/palemoon/locales/en-US/chrome/browser/browser.dtd
JustOff: No, `mediaLoop.accesskey` and `mediaPlaybackRate.accesskey` are on the same level, so I think it's better to rename `mediaPlaybackRate` to `mediaPlaybackRate2` (both `label` and `accesskey`) and change the `accesskey` from `l` to `d` for, as in Basilisk.

Ascrod: Oops, you're absolutely right. I completely misread which menu item you were referring to - will push a correction shortly.

mattatobin: Functionally speaking accesskeys only need to be unique on the same level.. Though, visually they MAY cause some confusion. I say change Ludicrous Speed access key to `u` instead.

JustOff: Yep, thanks!

Issue #971, opened 2019-02-13
browser.link.open_newwindow doesn't work properly for external links


wolfbeast:
This seems to be a Tycho -> UXP regression in the FE.
Looks like it may be related to the context now also being abused to pass "opener" flags from the back-end, making a simple comparison to known values fail.

Compare http://xref.palemoon.org/uxp-trunk/source/application/palemoon/base/content/browser.js#4411
and http://xref.palemoon.org/uxp-trunk/source/application/basilisk/base/content/browser.js#4741

wolfbeast:
PR #972 _should_ fix this. Landing this untested as an exception because of testing limitations in my work env. If people find it doesn't work while it should, we can re-open this issue.

wolfbeast:
Apparently this did not fix the issue. Reopening.

adeshkp:
This is not related to external links only and is reproducible with having something like `<a target="cool" href="http://palemoon.org/">PM</a>` in a page open in Pale Moon.

The issue here is that `browser.js`->`openURI` is called every time a link with a specific target is clicked in Pale Moon. It doesn't matter whether a window (tab) with the target name is already open in the browser or not.
As you know Basilisk is unaffected by this bug and it's behaviour is that it creates a new tab first time the link is clicked by calling openURI function. On subsequent clicks on the _same_ link (or different links with same target), it bypasses openURI call and loads the link directly in the already open tab.

My hunch is that Pale Moon is not able to query the browser to find the target window/tab, as it may be passing the wrong target name or not passing target at all and treating everything as `_blank`.

I was trying to fix this quickly and hoping to see this in 28.5.0, but now it's the end of the road for me. Don't know how to proceed further.

FranklinDM:
I think I've found the cause and solution to this issue.

When opening external links, an error is thrown to the console:
```
ReferenceError: can't access lexical declaration `referrer` before initialization  
browser.js:10975:37
	openURI chrome://browser/content/browser.js:10975:37
	handURIToExistingBrowser resource://app/components/nsBrowserContentHandler.js:687:3
	dch_handle resource://app/components/nsBrowserContentHandler.js:772:11
```
The problem seems to be the declaration of `referrer` here inside the switch block, case 2 (OPEN_NEWTAB):
https://github.com/MoonchildProductions/UXP/blob/41731a7f31a1724e74b32eb7be87de2719f977c3/application/palemoon/base/content/browser.js#L4484

[...]

PR #972, opened 2019-02-13
Treat aContext as a bitfield in openURI.

wolfbeast:
Potential fix for #971 

Just going by code comparison; I don't have an easy way to test this.


mattatobin:
The good news is if it is wrong it is STILL wrong but if it is fixed.. it is fixed.

wolfbeast:
I'm just going to merge this, build RCs and/or release and if it's not fixed then we can get back to this.

PR #985, opened 2019-03-04
Aboutbox redesign

### LINKED ISSUE #963: Redesign Pale Moon about box
The About box needs an overhaul.

- [x] Add "check for updates" to main menu and AppMenu so it can be removed from the about box.
- [x] Create new graphics for the about box
- [x] Redesign the content layout

wolfbeast:
Redesigns the about box:

- remove updater gunk
- replace background image with new style for official/unstable branding
- redesign the about box XUL layout and style

This resolves #963 

image1
image2


Nuck-TH:
Maybe i'm not expert on graphics design, but text is barely readable in both examples.
Current About window has relatively same brightness and color with smoth transitions under all text, so it reads good.
But here background is very contrast, which is bad for text readability.
Also in current window footer logically separated and contained license information, while now they just hand in abyss. However this one may be just my taste and i can't really recommend anything to improve it.

Lootyhoof:
The moon feels like it should align with the "End user rights" button since they're both mostly-centred, but they don't. And yet, the spacing on the buttons is fine as-is, so not really sure the best solution to that...

Just a thought is all since it irked me a little.

wolfbeast:
I've already added a "fog" behind the text to improve readability but it's not strong enough. At the same time I don't want to hide large portions of the news graphics. I'll tweak it some more. I do have an idea to solve this dilemma.

wolfbeast:
Added some animation to show the full graphic first then fade in the credits blurb, ending at this state:
about-updated1

wolfbeast:
Anyone have any further remarks? If not I'll merge this in

mrnhmath:
I'm afraid the proposed redesign looks busier and generally less streamlined than the current dialog.

mattatobin:
No comment. I guess.

JustOff:
Maybe it makes sense to try to increase the left and right indents of the main text, and center it?

mattatobin:
And don't use a blur on the edges.. Keep it consistant with the small links.

wolfbeast:
> I'm afraid the proposed redesign looks busier and generally less streamlined than the current dialog.

So does a Monet when compared to an empty wall.

It's an about box. It doesn't need to be streamlined or "clean" because all it does is provide some information and credits. I've used it to have some artistic expression in the browser and that is what this redesign is all about.

wolfbeast:
The "blur" was intended as a more subtle effect but with the heavier need for background color it doesn't work as I intended. I'll restyle it to be more of a text field with some centering.

wolfbeast:
This is where we are now. r+?
about-official2
about-unofficial2
about-unstable2

JustOff:
I definitely like these more, but, if I may, how about centering the moon vertically over the bottom info block and writing the name and version directly on it (also center vertically)?

wolfbeast:
> I definitely like these more, but, if I may, how about centering the moon vertically over the bottom info block and writing the name and version directly on it (also center vertically)?

Excuse me.. what?

JustOff:
53845248-ac57c680-3fa8-11e9-9c56-4418ad79dcc4

Sorry, I'm on VNC to my PC from mobile atm)

wolfbeast:
Oh, hell no. Sorry but I don't want text plastered -over- the moon; that's just tacky and so.. 90s. Also, the background is fixed artwork, I can't just move elements around.

JustOff:
It seems that these lines need to be removed

https://github.com/MoonchildProductions/UXP/blob/403605efb578527388f4af523eefaebaeb7e10ae/application/palemoon/base/content/aboutDialog.xul#L27-L29

as it results to `ReferenceError: onUnload is not defined`, because `onUnload()` no longer exists in `aboutDialog.js`.

wolfbeast:
Please don't comment on merged&closed pull requests. if it's related, comment on the relevant issue or create a new one.

JustOff:
Sorry, I usually follow this rule, but here I just confused the pull request with the issue.

PR #1023, opened 2019-03-25
Remove obsolete getPref helpers.

### LINKED ISSUE #989: Remove getPref helpers when they are obsolete.
With #988 implementing support for a default value for `get*Pref()`, the getPref helpers scattered in our tree will become obsolete and should be removed, and replaced with direct calls to `get*Pref()`

Depends on #988, so filing this as on hold until that is complete.

Ascrod:
Resolves #989. I think I got most of them, but let me know if there are any I missed.


mattatobin:
I wish you had not done it as one big commit..

PR #1035, opened 2019-04-03
Fix SVG mask crashes

### LINKED ISSUE #1034: SVG mask code can crash under certain conditions
Reported on the forum [here](https://forum.palemoon.org/viewtopic.php?f=37&t=21779). Similar crash was discovered and labeled as [Bug 1360343](https://bugzilla.mozilla.org/show_bug.cgi?id=1360343) by Mozilla. Basically the `maskSurface` can point to `nil` due to hardware condition and `maskFrame` can point to `nil` if the SVG contains invalid mask URL.

kn-yami:
This fixes #1034 by adding a `nullptr` check for both `maskSurface` and `maskFrame`.

### DIFF
--- a/gfx/2d/2D.h
+++ b/gfx/2d/2D.h
@@ -488,6 +488,9 @@ public:
   /**
    * Returns a DataSourceSurface with the same data as this one, but
    * guaranteed to have surface->GetType() == SurfaceType::DATA.
+   *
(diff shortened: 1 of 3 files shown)


kn-yami:
I see. I will remove it next time.

PR #1041, opened 2019-04-13
Clean up try/catch blocks for preferences

### LINKED ISSUE #991: Clean up try/catch blocks for preferences.
With #988 implementing defaults for preference getters, we should evaluate all instances of try/catch blocks for preferences since with a default value this no longer throws, and provide defaults where prudent.
Important: just blindly getting rid of these blocks is **not good** because in a good number of cases an "unset pref" should have different behavior and there is no "default value" to use.

Depends on #988 being implemented, so filing as `on hold`.

Ascrod:
Resolves #991.

This PR touches quite a lot. I tested some basic functionality as best I could, but there were a number of things that I either couldn't test or didn't know how. I'd appreciate a thorough review from anyone more familiar with some of this code than I.


devtools/client/framework/toolbox-process-window.js
```diff
@@ -100,17 +100,12 @@ function openToolbox({ form, chrome, isTabActor }) {
...
-    try {
-      // But if we are testing, then it should always open the debugger panel.
-      selectedTool = Services.prefs.getCharPref("devtools.browsertoolbox.panel");
-    } catch(e) {}
+    // Remember the last panel that was used inside of this profile.
+    // But if we are testing, then it should always open the debugger panel.
+    let selectedTool =
+      Services.prefs.getCharPref("devtools.toolbox.selectedTool",
+                                 Services.prefs.getCharPref("devtools.browsertoolbox.panel",
+                                                            "jsdebugger"));
```
JustOff: It seems to me you violated the original logic here.

toolkit/components/search/current/nsSearchService.js
```diff
@@ -993,19 +981,13 @@ function QueryParameter(aName, aValue, aPurpose) {
...
-      official = "unofficial";
-  }
-  catch (ex) { }
+  var distributionID = Services.prefs.getCharPref(BROWSER_SEARCH_PREF + "distributionID",
+                                                Services.appinfo.distributionID || "");
+  var official;
+  if (Services.prefs.getBoolPref(BROWSER_SEARCH_PREF + "official", MOZ_OFFICIAL))
+    official = "official";
+  else
+    official = "unofficial";
```
JustOff: I think this is wrong, at least because MOZ_OFFICIAL is not boolean.

toolkit/components/search/orginal/nsSearchService.js
```diff
@@ -1034,19 +1022,13 @@ function QueryParameter(aName, aValue, aPurpose) {
...
-      official = "unofficial";
-  }
-  catch (ex) { }
+  var distributionID = Services.prefs.getCharPref(BROWSER_SEARCH_PREF + "distributionID",
+                                                MOZ_DISTRIBUTION_ID || "");
+  var official;
+  if (Services.prefs.getBoolPref(BROWSER_SEARCH_PREF + "official", MOZ_OFFICIAL))
+    official = "official";
+  else
+    official = "unofficial";
```
JustOff: Same as above.

toolkit/content/widgets/scrollbox.xml
```diff
@@ -737,12 +727,9 @@
...
       <constructor><![CDATA[
-        try {
-          this._scrollDelay = this._prefBranch
-                                  .getIntPref("toolkit.scrollbox.clickToScroll.scrollDelay");
-        }
-        catch (ex) {
-        }
+        this._scrollDelay = this._prefBranch
+                                .getIntPref("toolkit.scrollbox.clickToScroll.scrollDelay",
+                                            this._scrollDelay);
```
JustOff: Are you sure that `this._scrollDelay` is always defined here?

devtools/client/framework/toolbox-process-window.js
Ascrod: Yep, the order of those two prefs got accidentally swapped. devtools.browsertoolbox.panel should take precedence.

toolkit/components/search/current/nsSearchService.js
Ascrod: You're right. I was following the Mozilla patches for this, but looking at it again it looks like their original patch is flawed and that the try-catch statement should remain.

toolkit/content/widgets/scrollbox.xml
Ascrod: There's a field definition below. If I'm not mistaken, this should always initialize the value to 150.
https://github.com/MoonchildProductions/UXP/blob/8757548a3e5b5cc4d159d61eb4a12c7af1be548e/toolkit/content/widgets/scrollbox.xml#L744

JustOff: Yeah, sorry I didn't notice.

JustOff:
LGTM otherwise.

mattatobin:
This is going to conflict with the sync-weave branch can you cross-reference and backout the Basilisk sync related changes until that lands?

wolfbeast:
If it's not actually touching the same code, it should be fine -- that's what merging is for after all ;-)
I'll give sync-weave priority here since that's a lot more involved -- if there are any merge conflicts afterwards, github will tell you and those can then be solved. backing out shouldn't be needed, just not going to merge this PR before the sync work.

mattatobin:
Understood.

wolfbeast:
@Ascrod the sync branch was merged with the fe changes and it's caused a few merge conflicts. If you'll solve those by fixing up your branch I'll merge this one in too.

Ascrod:
I'm on it.

toolkit/modules/UpdateUtils.jsm
```diff
@@ -31,13 +31,9 @@ this.UpdateUtils = {
    *        Whether or not to include the partner bits. Default: true.
    */
   getUpdateChannel(aIncludePartners = true) {
-    let channel = AppConstants.MOZ_UPDATE_CHANNEL;
+    let channel = defaults.getCharPref("app.update.channel",
+                                       AppConstants.MOZ_UPDATE_CHANNEL);
     let defaults = Services.prefs.getDefaultBranch(null);
```
JustOff: Simple, but unfortunate error: the variable `defaults` needs to be defined before use, and not vice versa.

PR #1045, opened 2019-04-20
Remove unused parameter from call

### LINKED ISSUE #1043: Minor annoyance in viewSource-content.js
This is in omi.ja!/chrome/toolkit/content/global/viewSource-content.js (which is a one to one copy of  toolkit/components/viewsource/content/viewSource-content.js from the repo).

at line 220 there is this definition

```
/**
   * Called when the parent sends a message to view some source code.
   *
   * @param URL (required)
   *        The URL string of the source to be shown.
   * @param outerWindowID (optional)
   *        The outerWindowID of the content window that has hosted
   *        the document, in case we want to retrieve it from the network
   *        cache.
   * @param lineNumber (optional)
   *        The line number to focus as soon as the source has finished
   *        loading.
   */
  viewSource(URL, outerWindowID, lineNumber) {
```

(3 params)

but it is called at line 130 with

```
    switch (msg.name) {
      case "ViewSource:LoadSource":
        this.viewSource(data.URL, data.outerWindowID, data.lineNumber,
                        data.shouldWrap);
```

(4 params)

So is passing a parameter which is just thrown away. Moreoever, if you have js strict mode warnings enabled, you get complaints every time you click 'view source in debugger' in the debug window that shouldWrap is an undefined property

ThosRTanner:
Removes an unnecessary parameter (which also references an undefined property) from the call to viewSource

Fixes #1043

### DIFF
--- a/toolkit/components/viewsource/content/viewSource-content.js
+++ b/toolkit/components/viewsource/content/viewSource-content.js
@@ -128,8 +128,7 @@ var ViewSourceContent = {
     let objects = msg.objects;
     switch (msg.name) {
       case "ViewSource:LoadSource":
-        this.viewSource(data.URL, data.outerWindowID, data.lineNumber,
-                        data.shouldWrap);
+        this.viewSource(data.URL, data.outerWindowID, data.lineNumber);
         break;
       case "ViewSource:LoadSourceOriginal":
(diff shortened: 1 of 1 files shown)


mattatobin:
Please don't submit Pull Requests without one of us commenting.

PR #1048, opened 2019-04-21
Remove geo-specific selection of default search engines.

### LINKED ISSUE #245: Remove geo-specific selection of default search engines.
This is a Mozilla-ism that defaults to probing a Mozilla server to get non-default server selections for search in an undocumented way for individual locales (the default way is apparently "not good enough" to have search engines and orders set in locale-specific properties).

Prefs are even used to set individual fallbacks in case the lookup fails and a custom list can't be grabbed from a URL.

Reference prefs: `browser.search.geoSpecificDefaults.*`, `browser.search.defaultenginename.{locale}` and `browser.search.order.{locale}.{n}`.

- [X] Disable geo-specific changes and the Mozilla service lookup for release
- [ ] Remove `browser.search.geoSpecificDefaults`
- [ ] Remove the relevant code of the search engine update client

All this needs to go.

Ascrod:
This resolves #245. I haven't modified the tests (which can probably be removed anyway) or anything in mobile that still uses the removed prefs.

@mattatobin indicated that he had a solution in mind for this... please let me know if there is anything I can do to improve this PR.

### DIFF
--- a/application/basilisk/app/profile/basilisk.js
+++ b/application/basilisk/app/profile/basilisk.js
@@ -346,18 +346,6 @@ pref("browser.search.order.1",                "chrome://browser-region/locale/re
 pref("browser.search.order.2",                "chrome://browser-region/locale/region.properties");
 pref("browser.search.order.3",                "chrome://browser-region/locale/region.properties");
 
-// Market-specific search defaults
-// This is disabled globally, and then enabled for individual locales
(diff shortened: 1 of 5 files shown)


mattatobin:
The code removal needs to happen but I don't know if can be done yet.

mattatobin:
r+

PR #1050, opened 2019-04-21
Fix issue with getUpdateChannel. Tag #991

Ascrod:
Follow up PR for a regression from #1041. See https://github.com/MoonchildProductions/UXP/pull/1041/files#r277165462.


Ascrod:
Oops, I didn't see Tobin's fix until just now. Sorry for the bugspam.

PR #1060, opened 2019-04-27
When using XRENDER, create Xlib-based offscreen image surfaces.

Ionic:
Mozilla tried to enable XRENDER support with backends other than Cairo
in 286348:e13aaaaf1962 /
https://hg.mozilla.org/mozilla-central/rev/e13aaaaf1962 - at least until
they decided to completely remove XRENDER support.

The change looked innocent enough, but actually turned out to do exactly
the opposite: it forcefully enabled image offscreen surfaces with GTK2
when it was previously disabled (since
gfxPrefs::UseImageOffscreenSurfaces() will always return true) and, by
extension, disabled the XRENDER-based functionality by creating a
non-Xlib surface.

Interestingly, a previously enabled double buffering check was also
disabled by this, but since the comment for this was diverging with the
code, that behavior just sounds like yet another bug.

Instead of disabling image offscreen surfaces (at least when using the
GTK2 backend), let's force the creation of Xlib-based image surfaces
when XRENDER support is enabled. This will let UXP use the more
common/modern code paths, but also make scrolling much faster again. Too
fast scrolling may induce tearing (if not smoothed), but on the other
hand performs much better in remote computing contexts.

As an added benefit, GTK3-based builds should roughly behave the same
way. Further tests with the GTK3-backend enabled will be required in the
future.

### DIFF
--- a/gfx/thebes/gfxPlatformGtk.cpp
+++ b/gfx/thebes/gfxPlatformGtk.cpp
@@ -156,7 +156,7 @@ gfxPlatformGtk::CreateOffscreenSurface(const IntSize& aSize,
     if (gdkScreen) {
         // When forcing PaintedLayers to use image surfaces for content,
         // force creation of gfxImageSurface surfaces.
-        if (gfxVars::UseXRender() && !UseImageOffscreenSurfaces()) {
+        if (gfxVars::UseXRender()) {
             Screen *screen = gdk_x11_screen_get_xscreen(gdkScreen);
             XRenderPictFormat* xrenderFormat =
                 gfxXlibSurface::FindRenderFormat(DisplayOfScreen(screen),
@@ -166,13 +166,6 @@ gfxPlatformGtk::CreateOffscreenSurface(const IntSize& aSize,
                 newSurface = gfxXlibSurface::Create(screen, xrenderFormat,
                                                     aSize);
             }
-        } else {
-            // We're not going to use XRender, so we don't need to
-            // search for a render format
-            newSurface = new gfxImageSurface(aSize, aFormat);
-            // The gfxImageSurface ctor zeroes this for us, no need to
-            // waste time clearing again
-            needsClear = false;
         }
     }
 #endif
@@ -182,6 +175,10 @@ gfxPlatformGtk::CreateOffscreenSurface(const IntSize& aSize,
         // e.g., no display, no RENDER, bad size, etc.
         // Fall back to image surface for the data.
         newSurface = new gfxImageSurface(aSize, aFormat);
+
+        // The gfxImageSurface ctor zeroes this for us, no need to
+        // waste time clearing again
+        needsClear = false;
     }
 
     if (newSurface->CairoStatus()) {


mattatobin:
The proper procedure is either to open discussion on the Pale Moon forums. At the very least open an issue so we can decide if this is something we want or need.

Drive-by pull requests are not very kindly taken here. But you have done it so let's move on.

@trava90

Ionic:
I probably should have mentioned some context: a few months ago, Stefan presented you with the fact that newer PM versions (28) behave much worse when scrolling (and probably also other fast-updating operations) than older ones (27-).

@wolfbeast, as far as I remember, was of the opinion that this regression isn't something that PM actually introduced, but more likely stem from Firefox (remember that PM 27 was based on FF-ESR 45, while PM 28 is based on FF-ESR 52). We all agreed on this - the general course of action everyone agreed one was that Firefox would have to be bisected to find out what exactly caused this performance regression.

This is what I did for the past few months - and also my findings.

mattatobin:
More like over a year now.. But yeah, we know the situation. However, that doesn't negate the fact that procedure needs to be observed and @trav90 needs to review this cause he is primarily dealing with the Linux.

However, it would help if you and @stefanbaur could do some more work yourselves. Pale Moon (as compiled by us and the official packagers) uses GTK2 and Basilisk uses GTK3.. We need to know of the impact on both as well as with or without XRENDER enabled.

We have to serve general configuration and builds above special and edge cases like @stefanbaur's remote access stuffs.

wolfbeast:
@mattatobin Most of this was discussed directly with Stefan in e-mail, so the normal procedure was side-stepped in this case. The "remote xrender" issue is actually a much more common case than you might think with X being used remotely in many situations, Linux being primarily a remotely operated server OS after all.
As for the timeline, it would have been great to see this sooner, of course, but it _is_ something that only people with the right setup could effectively research and fix.

@Ionic I'm happy to see you've been able to pinpoint the underlying cause (which to me sounds like your typical Mozilla bug, breaking perfectly-good functionality while trying to half-ass something, and then not fixing the broken stuff but rather removing it altogether for something different...). I'm assuming you've tested the result in your own builds before offering up the PR. Please let me know if you've tested both Pale Moon and Basilisk or not, and if not, could you make and test builds of both browsers, getting feedback on both GTK2 and GTK3 use, here?

wolfbeast:
Looking at the linked BZ bug and commit for it, it'd probably be a good idea to reintroduce some of those cairo checks for xrender usage too, considering people can use Skia by choice as a rendering back-end and that will likely break if trying to use Xlib surfaces (I think?).

wolfbeast:
If we need to discuss code mechanics or test results, please direct that to the issue #1061 instead of this PR, so we can keep things better organized.

stefanbaur:
>@wolfbeast: As for the timeline, it would have been great to see this sooner, of course, but it is something that only people with the right setup could effectively research and fix.

This, and of course, there's the "tragedy of the digital commons" - when software is available for free, getting people to pay for a bug hunt isn't exactly easy. Most are unwilling, and those that are willing, are barely able to chip in because they have limited financial resources themselves. And with a limited budget, you can't just throw an army of paid devs at the problem until it disappears.

gfx/thebes/gfxPlatformGtk.cpp
```diff
@@ -166,13 +166,6 @@ gfxPlatformGtk::CreateOffscreenSurface(const IntSize& aSize,
                 newSurface = gfxXlibSurface::Create(screen, xrenderFormat,
                                                     aSize);
             }
-        } else {
-            // We're not going to use XRender, so we don't need to
```
wolfbeast: Don't we still need this block for when xrender is not in use (i.e. user-disabled)?

wolfbeast: Oh, never mind it will fall through with effectively the same result. Sorry for the noise.

Ionic: Right, I just merged it down. Retrying the same failed operation twice shouldn't lead to a different result. I also noticed that we can omit clearing the surface in the general case (which is also indicated by the comment), so I went and applied that to the general case as well. :)

PR #1074, opened 2019-05-04
Removed the telemetry folder that is in the toolkit/components folder 

softwaretechone:
This helps to resolve issue #21 

The purpose of this pull request is to remove the telemetry folder that is located in the toolkit/components folder.

I achieved this by first deleting all the files in the toolkit/components/telemetry folder. Second, of all the files in the folder, Telemetry.h was the most widely utilized file. That is why a large number of the commits mention Telemetry.h. Third, I deleted all the telemetry/health report files that were not in the toolkit/components/telemetry folder but were dependent to files within the folder. Fourth, I cleaned up any leftoever telemetry/health reporting/crash reporting code from the code base that utilized the files in the telemetry folder. Also, I deleted any telemetry/health reporting/crash reporting code that might not have used the files in the telemetry folder but was tied together with telemetry/health reporting/crash reporting code that was. Finally, the successful removal of the telemetry folder was due to removing the folder as a whole and not by increments due to how closely tied all the telemetry code was.


mattatobin:
This is still unacceptable.. Please check commit history to see how we handle removals such as this.

If you have any questions, please comment in the orginal issue so that you know how things should be done before you waste effort doing it improperly.

PR #1080, opened 2019-05-05
Reader Updates

Ascrod:
This update includes changes to the Readability library ported from Firefox 66.0.3, as well as a handful of other bugfixes. I've held off on including [Bug 1187696](https://bugzilla.mozilla.org/show_bug.cgi?id=1187696) since I haven't quite figured out how to tackle it yet.

The Readability library update is a breaking change, so any external consumers of the module (e.g. Reader View) may need to be updated to accommodate it.

As I mentioned in the comments for #361, this update also includes some new tooltips that will need translations.


JustOff:
String changes were pushed to CrowdIn.

PR #1085, opened 2019-05-10
Add null check for URL Bar in the writePlaceholder() method

### LINKED ISSUE #1084: Add null check for URL Bar in the writePlaceholder() method
As [reported](https://forum.palemoon.org/viewtopic.php?f=29&t=22072&p=166725) on the forum, the browser interface (including places and others) becomes broken when the URL bar is removed.

issue new

The culprit seems to be a missing null check for `gURLBar` at the beginning of the writePlaceholder() method in [line 414](http://xref.palemoon.org/uxp-trunk/source/application/palemoon/base/content/browser.js#414) of `browser.js`.

FranklinDM:
This resolves #1084.
Does as the title says, adds a null check for the URL Bar in the writePlaceholder() method.

### DIFF
--- a/application/palemoon/base/content/browser.js
+++ b/application/palemoon/base/content/browser.js
@@ -412,6 +412,9 @@ var gURLBarSettings = {
   },
 
   writePlaceholder: function() {
+    if (!gURLBar)
+      return;
+
     let attribute = "placeholder";
     let prefs = this.prefSuggests.map(pref => {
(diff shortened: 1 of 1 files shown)


mattatobin:
We prefer conditionals to have braces but don't worry about it this time unless you want to. Just remember for the future.

I will function test this later today so stand by.

Issue #1090, opened 2019-05-17
Remove unboxed arrays


wolfbeast:
Similar to #1030, but with the difference that this code was never even enabled on release and is just dead code sitting there and adding unneeded complexity to SpiderMonkey/TraceMonkey/IonMonkey, as well as complicating the removal in #1030 by still requiring functions to remain in place for unboxed arrays.

Bug to use as guideline: https://bugzilla.mozilla.org/show_bug.cgi?id=1398768

Issue #1030, opened 2019-03-28
Remove "unboxed" typed objects


wolfbeast:
Follow-up to #1017 

Disabling the "unboxed" object optimization gives us better performance in the real world. This has been baking without ill effects on unstable so we can remove the code now, simplifying typed object handling and reducing quite a bit of (complex) code in SpiderMonkey as a result.

We can use https://bugzilla.mozilla.org/show_bug.cgi?id=1505574 as a guideline.

mattatobin:
Might be wise to let this sit for a release cycle so that it can get broad usership before we kill something we may have to put back in.

wolfbeast:
Oh I agree with keeping the pref-controlled state for a while. No rush on this.

mattatobin:
Planned removal for Pale Moon 28.5.1 and matched UXP Applications.

wolfbeast:
There have been 0 reports of issues disabling these objects, so let's go ahead and remove this code.
@mattatobin are you taking this or shall I? Looks like this is at least something I won't brickwall with this week ;P

mattatobin:
Have a blast with it.

wolfbeast:
This is a lot of cleanup work. Going to take a while removing this and testing inbetween chunks.

wolfbeast:
Depends on #1090. I'll have to split this issue up so I can remove arrays before continuing removing this.

wolfbeast:
Because this is more work than expected, I'm pushing this back to 28.6

g4jc:
FYI, there's a linker error introduced after the merge of Phase 1:

```
/js/src/jit/IonCaches.cpp:1545: error: undefined reference to 'js::UnboxedPlainObject::layout() const'
/js/src/jit/IonCaches.cpp:3414: error: undefined reference to 'js::UnboxedPlainObject::layout() const'
/js/src/jit/Recover.cpp:1543: error: undefined reference to 'js::UnboxedPlainObject::layout() const'
collect2: error: ld returned 1 exit status
```

adeshkp:
@g4jc I encountered this on Mac few days ago. Linux build was somehow unaffected, so didn't bother. It can be remedied by including "vm/Unboxed-inl.h" header file.

adeshkp:
@g4jc I encountered this on Mac few days ago. Linux build was somehow unaffected, so didn't bother. It can be remedied by including "vm/Unboxed-inl.h" header in the mentioned files.

[...]

PR #1091, opened 2019-05-17
Remove unboxed objects phase 1

wolfbeast:
This removes a good chunk of unboxed object code, but will have to be done in two phases to slide #1090 in-between before I continue.

Sequence:
1. This PR, phase 1
2. Remove unboxed arrays
3. Revisit #1030 and double-check code removal, then finish up in phase 2


wolfbeast:
Could use a code review from someone familiar with C++, just in case I made typoes/mistakes.

Issue #1061, opened 2019-04-27
Fix XRENDER bottleneck


wolfbeast:
This should have been created when the issue was first brought up that, although xrender works (#341 ) , it suffers from a serious bottleneck.

Most of the discussion about this was communicated in private e-mail with @stefanbaur which ended up basically with "we'll have someone investigate the cause and come up with a PR" which took a few months.

It was found that the performance issue was created by fully redrawing scrolled surfaces after a delta is sent through via remote X. It wasn't until the PR was offered to our repo that we were informed about the actual issue underlying this. See also the information on the PR (#1060 )

wolfbeast:
Copying the explanation from the PR ( @Ionic ):

Mozilla tried to enable XRENDER support with backends other than Cairo
in 286348:e13aaaaf1962 /
https://hg.mozilla.org/mozilla-central/rev/e13aaaaf1962 - at least until
they decided to completely remove XRENDER support.

The change looked innocent enough, but actually turned out to do exactly
the opposite: it forcefully enabled image offscreen surfaces with GTK2
when it was previously disabled (since
gfxPrefs::UseImageOffscreenSurfaces() will always return true) and, by
extension, disabled the XRENDER-based functionality by creating a
non-Xlib surface.

Interestingly, a previously enabled double buffering check was also
disabled by this, but since the comment for this was diverging with the
code, that behavior just sounds like yet another bug.

Instead of disabling image offscreen surfaces (at least when using the
GTK2 backend), let's force the creation of Xlib-based image surfaces
when XRENDER support is enabled. This will let UXP use the more
common/modern code paths, but also make scrolling much faster again. Too
fast scrolling may induce tearing (if not smoothed), but on the other
hand performs much better in remote computing contexts.

As an added benefit, GTK3-based builds should roughly behave the same
way. Further tests with the GTK3-backend enabled will be required in the
future.

wolfbeast:
@Ionic Looking at the linked BZ bug and commit for it, it'd probably be a good idea to reintroduce some of those cairo checks for xrender usage too, considering people can use Skia by choice as a rendering back-end and that will likely break if trying to use Xlib surfaces (I think?).

Ionic:
[...]

PR #1100, opened 2019-05-22
Enable double buffering when using XRENDER.

Ionic:
This fixes flickering/bars/stripes showing up during quickly-updating
operations on Intel hardware when using XRENDER.

For more information, refer to the code comment.

See #1061

### DIFF
--- a/gfx/layers/client/ContentClient.cpp
+++ b/gfx/layers/client/ContentClient.cpp
@@ -78,8 +78,21 @@ ContentClient::CreateContentClient(CompositableForwarder* aForwarder)
   // We can't use double buffering when using image content with
   // Xrender support on Linux, as ContentHostDoubleBuffered is not
   // suited for direct uploads to the server.
+  // FIXME: Even though the comment above suggests that double buffering
+  //        is supposed to be disabled when Xrender support is being enabled
+  //        (and used), it really wasn't. Historically,
+  //        UseImageOffscreenSurfaces() was always false in GTK2 builds, thus
+  //        triggering the check, regardless of UseXRender().
+  //        Some time later, offscreen surfaces were always enabled, but the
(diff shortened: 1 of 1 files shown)


gfx/layers/client/ContentClient.cpp
```diff
@@ -78,8 +78,21 @@ ContentClient::CreateContentClient(CompositableForwarder* aForwarder)
...
+  //        Xrender functionality broke due to not using Xlib-based surfaces.
+  //        Using Xlib-based surfaces compatible with Xrender operations seems
+  //        to lead to weird graphical artifacts (bars and stripes) on some
+  //        hardware (Intel-based?) when displaying quickly-changing content,
+  //        so contrary to the statement above we'd better enable double
+  //        buffering - which also seems to not have any negative performance
+  //        impact.
   if (!gfxPlatformGtk::GetPlatform()->UseImageOffscreenSurfaces() ||
-      !gfxVars::UseXRender())
+       gfxVars::UseXRender())
```
wolfbeast: I didn't think about this previously, but shouldn't this xrender check simply be removed? We'd want to use double buffering regardless of the xrender setting (as long as off-screen surfaces aren't used), after all. What's happening now is that it would not be used if xrender is off, with this change. I think that's incorrect.

stefanbaur: Hi, @Ionic is having some technical difficulties with his machine, so he has asked me to reply on his behalf. Here's what he asked me to relay:
1.) The condition is "OR", not "AND", so even if xrender is off, we will use double buffering if offscreen surfaces are off.
2.) We want to always use double buffering with xrender.
-Stefan

wolfbeast: Right, my bad. Lack of sleep and all. Sorry for the noise.

Issue #1101, opened 2019-05-22
Web-Compat: Support gzip-compressed SVGs in OpenType+SVG fonts


g4jc:
Looks like a pretty straight forward implementation should we decide to support. Allows OpenType fonts to be rendered that have gzip compression. (ref: [m-c 1359240](https://bugzilla.mozilla.org/show_bug.cgi?id=1359240) Several font families are supporting this.

"""
While SVG 1.1 requires in addition to plain text encoding that conforming SVG implementations must correctly support gzip-encoded [RFC1952] and deflate-encoded [RFC1951] data streams, this specification requires that conforming SVG implementations must correctly support plain-text and gzip-encoded [RFC1952] data streams only. In both cases, svgDocLength encodes the length of the encoded data, not the decoded document.
"""

Patch: https://paste.debian.net/plainh/6943b234

PR #1108, opened 2019-05-23
Support gzip-compressed SVGs in OpenType+SVG fonts

g4jc:
Check if the SVG document is gz-compressed, and decompress before attempting to parse. RefTest included.


wolfbeast:
This resolves #1101

Issue #1102, opened 2019-05-22
Web-Compat: Disable <meta http-equiv=set-cookie>


g4jc:
This particular code is more or less a follow up to [Implement "cookie-averse document objects". ](https://github.com/MoonchildProductions/UXP/commit/85083fce2da7a270e324fd951b7f3d03a50aef1b#diff-df2e2c6060983ece1c6e0999fdb66025) which we already implemented. Since `<meta http-equiv=set-cookie>` is a non-standard way of setting cookies and has been removed from web standards, we can disable it and stuff it behind a pref. This was done in [m-c 1457503](https://bugzilla.mozilla.org/show_bug.cgi?id=1457503), but uses their `StaticPrefList.h` rather than our `modules/libpref/init/all.js`.

Preliminarily, this looks pretty easy to implement, and it doesn't fully remove the code if someone desperately needs it (which is highly unlikely since it's non-standard), so they can just flip the pref.

Patch: https://paste.debian.net/plainh/9547a6d2
_Note: Need to make sure my pref and true/false parameters are correct as the implementation does not use StaticPrefList. So far it is build passing._

wolfbeast:
This isn't in ESR-60 but hopefully by now the world will have adjusted to stop using these. The pref is a good idea so we can easily recover if things go unexpectedly badly on release.

Please avoid static preference definitions at all costs in our tree. Always use the normal greprefs/app prefs mechanism.

g4jc:
Confirmed working on http://w3c-test.org/cookies/meta-blocked.html.
Pref is defaulted to: `dom.metaElement.setCookie.allowed;false`

g4jc:
Pref name changed to `dom.meta-set-cookie.enabled` [per request](https://github.com/MoonchildProductions/UXP/pull/1114#issuecomment-495708040).

PR #1114, opened 2019-05-24
Disable <meta http-equiv=set-cookie>

g4jc:
Sets a new preference: `dom.meta-set-cookie.enabled` to `false`, by default. Fixing #1102.

### DIFF
--- a/dom/base/nsContentSink.cpp
+++ b/dom/base/nsContentSink.cpp
@@ -304,7 +304,8 @@ nsContentSink::ProcessHeaderData(nsIAtom* aHeader, const nsAString& aValue,
 
   mDocument->SetHeaderData(aHeader, aValue);
 
-  if (aHeader == nsGkAtoms::setcookie) {
+  if (aHeader == nsGkAtoms::setcookie &&
+      Preferences::GetBool("dom.meta-set-cookie.enabled", true)) {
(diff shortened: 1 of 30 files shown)


mattatobin:
As far as prefs like this is conserned the only pref naming convention is from dom.meta-viewport.enabled..

Can we have the pref named dom.meta-setCookie.enabled instead?

What is your opinion, @wolfbeast?

wolfbeast:
I agree that the current pref name isn't really good. Let's make it `dom.meta-set-cookie.enabled` as a logical name for enabling setting of cookies via meta.

As for false by default, I think we're good on that.

mattatobin:
I prefer my pref name better but I am fine with Moonchild's. Anything is better than what I assume Mozilla used.

g4jc:
Pref name changed as requested. Not sure why github closed the PR?

g4jc:
Confirming bug in [#361 Github](https://github.com/isaacs/github/issues/361). TIL. Opening a new PR.

wolfbeast:
Force-pushing branches overwriting history makes things really hard to follow and requires pretty much a re-review of the entire PR after it. I'd prefer if you didn't do so in the future and just add changes as commits to the existing branch the PR is from -- they will automatically be picked up as part of the PR.

wolfbeast:
This resolves #1102

Issue #1111, opened 2019-05-24
NodeInfoManager should use a local cache


wolfbeast:
`NodeInfoManager` should use a local cache for recently used nodeinfos to avoid slow hashtable lookups.

https://bugzilla.mozilla.org/show_bug.cgi?id=1352235

**Note**: patch for this issue on BMO v1 also has a fix as a follow-up for #1110 which we would likely want (an atom lookup call being made cached).

Blocks #1109

PR #1115, opened 2019-05-24
Add NS_Atomize cache for main thread use

### LINKED ISSUE #1110: Add NS_Atomize cache for main thread use
Have a (static) HTML-parser specific hash lookup table which holds owning references to atoms for all HTML element/attribute names. That hash table can be fully lock-free since it's immutable after initial creation. We would have to fall back to doing a normal `NS_Atomize` if the parser-specific hash table doesn't find an atom since the page might be using a non-standard element/attribute. But that should be a much more rare case.

https://bugzilla.mozilla.org/show_bug.cgi?id=1351303

Blocks #1109

win7-7:
This adds add main thread only cache for nsIAtoms to speed up atomization. 

Based to https://bugzilla.mozilla.org/show_bug.cgi?id=1351303

This resolves #1110

### DIFF
--- a/dom/base/Element.cpp
+++ b/dom/base/Element.cpp
@@ -1316,7 +1316,7 @@ Element::SetAttribute(const nsAString& aName,
   nsAutoString nameToUse;
   const nsAttrName* name = InternalGetAttrNameFromQName(aName, &nameToUse);
   if (!name) {
-    nsCOMPtr<nsIAtom> nameAtom = NS_Atomize(nameToUse);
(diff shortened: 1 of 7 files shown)


mattatobin:
@win7-7 Please in the future check with me for commit text.

wolfbeast:
> This need be merged before issue #1111 can be done as Moonchild asked to v1 of that patch which depends on this.

I think you misunderstood. I wanted you to have a look at v1 of that patch and pick up the relevant change for _this_ PR.

win7-7:
Is this addition what you meant? Is this ready to be merged now?

PR #1141, opened 2019-06-26
[Pale Moon] [Linux] Fix for duplicate icon in GNOME app bar

kbhasi:
Fix is based on [this tutorial](https://www.youtube.com/watch?v=xZ_M3Q-U-J0).

PM provides 2 window classes, "Navigator" and "pale moon". I chose the former as I believe both official and unofficial branding packages have different classes for the latter, but the former should be identical between the two.

I think this will also fix broken startup notifications for PM in Plasma and Trinity, but I haven't checked to see if that is the case, as of filing this pull request.


wolfbeast:
Is there an issue this belongs to?

kbhasi:
Oh! Right. Yeah, there isn't.

I was the only one that noticed this issue, but as I was looking for a fix for the same issue with some KDE applications (before finding that they were fixed upstream, but not in Debian 10), I remembered that I had the same issue with PM.

kn-yami:
> Should I create an issue for this?

You probably should: https://github.com/MoonchildProductions/UXP/blob/master/docs/CONTRIBUTING.md#general-guideline-for-submitting-pull-requests

kbhasi:
Unfortunately, I'll have to close this PR as I noticed an error, as I should've known all Mozilla-based web browsers share that common "Navigator" window class.

PR #1144, opened 2019-06-27
Fix linking on Linux and Mac OS

trav90:
Just what it says on the tin.


mattatobin:
That's it?

wolfbeast:
> That's it?

Most deprot fixing is minimal. It's just figuring out what to place where that's the issue.

PR #1142, opened 2019-06-26
Remove unboxed objects

wolfbeast:
This removes the rest of the disabled unboxed objects code (phase 2). Like with unboxed arrays, this could do with a check-build on Mac and Linux before merging. 

This resolves #1030 :tada:

PR #1145, opened 2019-06-29
Issue #1142 - Follow up: Remove uneeded assertion

g4jc:
Since unboxed objects were removed, isUnboxed is undefined. This causes a build failure when debuging is enabled. This patch fixes the issue by removing the uneeeded assertion.


g4jc:
[For reference](https://hg.mozilla.org/mozilla-central/rev/136b21be1299f0ee657a4e6c4e95e626f706ea78).

mattatobin:
r+

Issue #1148, opened 2019-06-30
Duplicate icon + incorrectly displayed app name in GNOME, and startup feedback is broken


kbhasi:
I checked, and have traced the problem to not having a `StartupWMClass` property in the main `.desktop` file dropped in the user-space "Applications" folder.

**I already have a fix for this which I'll file a pull request for in a few minutes, hours, or days.**
Edit: While I do have a forked repo, I've confirmed the 2 commits I done in there to not fix this issue, and instead cause more conflicts (with other Mozilla-based web browsers).

1. Duplicate icon in GNOME
![Screenshot from 2019-06-30 17-17-56 (2)](https://user-images.githubusercontent.com/37286461/60394787-07b35300-9b5c-11e9-897e-806d29c24115.png)

2. Broken startup feedback (shown here in GNOME, but also exists in Plasma and Trinity)
![Screenshot from 2019-06-27 22-57-35 (2)](https://user-images.githubusercontent.com/37286461/60394796-24e82180-9b5c-11e9-9674-ee61f9af659f.png)

3. Incorrect capitalisation in GNOME in conjunction with the duplicate icon, where it appears to be going by the WM class.
![Screenshot from 2019-06-30 17-18-02 (2)](https://user-images.githubusercontent.com/37286461/60394821-8e683000-9b5c-11e9-857f-b2734e2ccebf.png)

kbhasi:
I've fixed points 1 and 3 by adding this line at line 92:

```.desktop
StartupWMClass="pale moon"
```

However, I haven't fixed point 2, as I tried to do so, but it seems UXP may not be sending the ["remove" signal](https://cgit.freedesktop.org/startup-notification/tree/doc/startup-notification.txt#n203) once the UI loads, first page finishes loading, or page loading is stopped.

The easy way out of it, though, would be to do this to line 90 of the file:

```.desktop
StartupNotify=false
```

Should I do this, or will there be specific GNU/Linux (and BSD?) fixes done to somewhere else on the platform itself to allow for the proper signal to be sent?

trav90:
Startup notification support for UXP applications is disabled by default at build time (unless explicitly enabled in .mozconfig), so setting StartupNotify to false would be the correct course until/if we do start building startup notification support.

Please submit a new PR and I will review it.

kbhasi:
@trav90 Great! I'll take the easy way out. PR coming later tonight.

kbhasi:
[...]

PR #1149, opened 2019-07-01
Fix duplicate icons and incorrect capitalisations in GNOME, and disabled startup feedback

kbhasi:
Fix is based on [this tutorial](https://www.youtube.com/watch?v=xZ_M3Q-U-J0).

More info in issue #1148 .


trav90:
r+

PR #1151, opened 2019-07-02
 Attach FrameProperties to each frame instead of using a shared hashtable

### LINKED ISSUE #1150: Attach FrameProperties to each frame instead of using a shared hashtable
main patch:

https://bugzilla.mozilla.org/show_bug.cgi?id=1365982

For example, in https://perfht.ml/2rv39uP, which is loading the single-page HTML spec, there is 355ms spent in FramePropertyTable methods; around half of this is spent in PLDHashTable methods, i.e. looking up the frame pointer in the hashtable to find its list of properties.

What I'd like to consider here is dispensing with the shared hashtable and instead attaching the frame property list directly to nsIFrame. This will add one pointer to the size of all frame classes, but will completely eliminate the hashtable cost here.

This improves http://sinz.org/Maze/ along with some general performance improvements.

related patches that would be included in pull request:

https://bugzilla.mozilla.org/show_bug.cgi?id=1378219 (fixes crash)

https://bugzilla.mozilla.org/show_bug.cgi?id=1368654 (memory reporter)

https://bugzilla.mozilla.org/show_bug.cgi?id=1368369 (reduce RAM usage)

win7-7:
This pull request is based to: 

main patch:

https://bugzilla.mozilla.org/show_bug.cgi?id=1365982

and related patches:

https://bugzilla.mozilla.org/show_bug.cgi?id=1378219 (fixes crash)

https://bugzilla.mozilla.org/show_bug.cgi?id=1368654 (memory reporter)

https://bugzilla.mozilla.org/show_bug.cgi?id=1368369 (reduce RAM usage)

This PR has changes into FrameLayerBuilder.cpp and nsTableFrame.cpp and .h files to get main patch and related patches to compile with UXP but works after those changes without noticeable problems during usage.

Resolves #1150


wolfbeast:
This touches a lot of different components... Have you thoroughly tested these changes, including e.g. SVG effects and the like?

win7-7:
Tested many SVG demos for gradients, filter effects, pattern, clipping and masking and it looks same as in Basilisk when compared.

wolfbeast:
Fantastic!

PR #1155, opened 2019-07-03
Look into optimizing out the hashtable lookups from nsContainerFrame::DestroyFrom and nsContainerFrame::SafelyDestroyFrameListProp

### LINKED ISSUE #1154: Look into optimizing out the hashtable lookups from nsContainerFrame::DestroyFrom and nsContainerFrame::SafelyDestroyFrameListProp
https://bugzilla.mozilla.org/show_bug.cgi?id=1367206

I think all these child list properties are rare so iterating over
the properties once upfront instead of four separate lookups
should be a win. Now that  #1151 is merged this can be done next.

This further improves results of:

http://sinz.org/Maze/

win7-7:
This further improves performance after merging  #1151 to UXP.

Resolves #1154.

### DIFF
--- a/layout/base/FrameProperties.h
+++ b/layout/base/FrameProperties.h
@@ -161,6 +161,11 @@ public:
   }
 
   /**
+  * Return true if we have no properties, otherwise return false.
+  */
+ bool IsEmpty() const { return mProperties.IsEmpty(); }
+
+  /**
    * Set a property value. This requires a linear search through
    * the properties of the frame. Any existing value for the property
    * is destroyed.
@@ -241,6 +246,26 @@ public:
   {
     DeleteInternal(aProperty, aFrame);
   }
+  
+   /**
+   * Call @aFunction for each property or until @aFunction returns false.
+   */
+  template<class F>
+  void ForEach(F aFunction) const
+  {
+#ifdef DEBUG
+   size_t len = mProperties.Length();
+#endif
+    for (const auto& prop : mProperties) {
+      bool shouldContinue = aFunction(prop.mProperty, prop.mValue);
+      MOZ_ASSERT(len == mProperties.Length(),
+                 "frame property list was modified by ForEach callback!");
+      if (!shouldContinue) {
+        return;
+      }
+    }
+  }
+
   /**
    * Remove and destroy all property values for the frame.
    */
--- a/layout/generic/nsContainerFrame.cpp
+++ b/layout/generic/nsContainerFrame.cpp
@@ -219,24 +219,49 @@ nsContainerFrame::DestroyFrom(nsIFrame* aDestructRoot)
   // Destroy frames on the principal child list.
   mFrames.DestroyFramesFrom(aDestructRoot);
 
+
+  if (MOZ_UNLIKELY(!mProperties.IsEmpty())) {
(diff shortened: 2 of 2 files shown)


layout/base/FrameProperties.h
```diff
@@ -241,6 +246,26 @@ class FrameProperties
...
   }
+  
+   /**
+   * Call @aFunction for each property or until @aFunction returns false.
+   */
+  template<class F>
+  void ForEach(F aFunction) const
+  {
+#ifdef DEBUG
+   size_t len = mProperties.Length();
```
wolfbeast: Wrong indent.
Also, only defining this in debug but using it in an assert outside of debug is bad, see line 261.

layout/generic/nsContainerFrame.cpp
```diff
@@ -219,24 +219,49 @@ nsContainerFrame::DestroyFrom(nsIFrame* aDestructRoot)
...
-  MOZ_ASSERT(IsFrameOfType(nsIFrame::eCanContainOverflowContainers) ||
-             !(GetProperty(nsContainerFrame::OverflowContainersProperty()) ||
-               GetProperty(nsContainerFrame::ExcessOverflowContainersProperty())),
-             "this type of frame should't have overflow containers");
-  SafelyDestroyFrameListProp(aDestructRoot, shell,
-                             OverflowContainersProperty());
-  SafelyDestroyFrameListProp(aDestructRoot, shell,
-                             ExcessOverflowContainersProperty());
+  MOZ_ASSERT(IsFrameOfType(eCanContainOverflowContainers) ||
+                           !(hasOC || hasEOC),
```
wolfbeast: This is a very confusing logical operation; please rewrite to something more sensible.

wolfbeast:
Also, please make sure you use proper descriptive commit messages and pull request titles/descriptions -- as-is it's not acceptable because it will be impossible to re-find messages like this in the commit history on keywords.

win7-7:
Is this correct way to do first change requested/ defining?

   template<class F>
   void ForEach(F aFunction) const
   {
#ifdef DEBUG
      size_t len = mProperties.Length();
#endif
       for (const auto& prop : mProperties) {
          bool shouldContinue = aFunction(prop.mProperty, prop.mValue);
          size_t len = mProperties.Length();
          MOZ_ASSERT(len == mProperties.Length(),

Sorry I do not have proper knowledge how to fix these requested changes. There is nothing similar from Mozilla that could be used as to fix these parts. Mozilla files of these still use these same way as in this PR.

Could you also tell exactly how second change you requested should be done?

wolfbeast:
Look at your code
```
#ifdef DEBUG
size_t len = mProperties.Length();
#endif
```
len is only defined if you are in debug mode
A few lines below:
```
MOZ_ASSERT(len == mProperties.Length()
```
Using len in a check that is *not* in debug mode, so `len` will be undefined.
Just adding a definition right above it is pointless since the assertion will then always be true. The point is to check inside the loop whether `mProperties.Length()` has changed from the previous `len`.
Just wrap the assert in `#ifdef DEBUG` PP statements.

The second one is fine as-is, I guess. Not happy with the logic sequence but it works.

win7-7:
I will close this pull request to do new pull request instead.

PR #1174, opened 2019-07-11
Report matroska containers as supported and also support h.264 content in a matroska container file

JeroenVreeken:
See also

https://forum.palemoon.org/viewtopic.php?f=5&t=22516&sid=d29d299758de3bf84d59587bfaa1a213


wolfbeast:
LGTM

Issue #1156, opened 2019-07-05
Location Bar Preferences Checkboxes


RealityRipple:
The History, Bookmarks, and Open tabs checkboxes in the Location Bar Privacy Preferences behave incorrectly. If you uncheck all of them and then check any one of the three, all three will become checked. Tested with a blank profile, with and without saving/closing the Preferences between unchecking and checking. Possibly a bitwise flag issue, if I had to guess?

wolfbeast:
PR #1200 fixes this behavioral issue in a normal pref-apply situation. It's unknown whether it still toggles the `autocomplete.enabled` pref properly when instantApply is enabled (e.g. on Mac).

Ascrod:
This issue doesn't manifest at all on the Linux release build, which also uses `instantApply` by default. I can build from trunk today and test on Linux as well as Windows with `instantApply` enabled.

Ascrod:
LGTM in both cases. `browser.urlbar.suggest.*` and `browser.urlbar.autocomplete.enabled` appear to be set properly.

wolfbeast:
Fantastic! Just superfluous introduced code then.

PR #1200, opened 2019-07-27
Issue #1156 - Location Bar Preferences Checkboxes

flewkey:
Removes some unnecessary preference synchronizing functions from the privacy preferences pane.
The browser.url.autocomplete.enabled and browser.url.suggest.* preferences are already synced elsewhere.
This stops the pane from unnecessarily toggling browser.url.autocomplete.enabled and enabling all other browser.url.suggest.* preferences.
Fixes Issue #1156

### DIFF
--- a/application/palemoon/components/preferences/privacy.js
+++ b/application/palemoon/components/preferences/privacy.js
@@ -298,36 +298,6 @@ var gPrivacyPane = {
 
   // HISTORY
 
-  /**
-   * Read the location bar enabled and suggestion prefs
-   * @return Int value for suggestion menulist
-   */
-  readSuggestionPref: function PPP_readSuggestionPref()
-  {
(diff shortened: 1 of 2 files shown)


wolfbeast:
Where exactly is `browser.url.autocomplete.enabled` synced with these preferences "elsewhere"?

wolfbeast:
Looks good functionally -- it synchronizes on OK; I'm not sure about behavior on Mac though because that uses instantapply, which might not get triggered now.

flewkey:
> 
> 
> Where exactly is `browser.url.autocomplete.enabled` synced with these preferences "elsewhere"?

Apologies for the late response and for the vague PR message. Those preferences seem to be synced in nsPlacesAutoComplete.js, and removing the relevant code breaks syncing as expected.

PR #1205, opened 2019-08-08
Convert dom/base/nsImageLoadingContent.cpp to use AsyncOpen2 and followups along with it (1445670 and 1373780 part 2 and 3)

### LINKED ISSUE #1204: Convert dom/base/nsImageLoadingContent.cpp to use AsyncOpen2 and followups
As followup to Mozilla work in https://bugzilla.mozilla.org/show_bug.cgi?id=1206961, pull request that would solve this will convert nsImageLoadingContent.cpp to use AsyncOpen2() and also includes followups to main patch (1267075).

https://bugzilla.mozilla.org/show_bug.cgi?id=1267075 (main patch) (part 1,2,3 and 4 / without adding tests)

Part 1: call SetBlockedRequest when CSP check failed.

Remove calling CanLoadImage. Also in the case of CSP check failed, we will call SetBlockedRequest in those cases.

Part 2: bail out early if it's for data document.

Check the document earlier before calling into imageLoader, this could save us tons of time if in the end the document isn't allowed to load images.

Part 3: cancel imgRequestProxy if asyncOpen2 failed

The problem is if we found a cache hit, then we could go through ValidateRequestWithNewChannel to validate the cache. Then if the CSP check fail(asyncOpen2() will fail), then the imgRequestProxy will remain there, and cause the timeout.

Part 4: add a boolean to prevent calling asynchronously

Use a boolean to prevent calling SetBlockedRequest asynchronously. Also use the same boolean to prevent some evil code reentering LoadImage.

https://bugzilla.mozilla.org/show_bug.cgi?id=1373780 (part 2 and 3)

Part 2: remove calling nsContentUtils::CanLoadImage().

Part 3: add isImgSet argument.

https://bugzilla.mozilla.org/show_bug.cgi?id=1445670

 Make sure to clear out the pending request when our image gets blocked by the data document policy.

win7-7:
As followup to Mozilla work in https://bugzilla.mozilla.org/show_bug.cgi?id=1206961, this will convert nsImageLoadingContent.cpp to use AsyncOpen2() and also includes followups to main patch.

https://bugzilla.mozilla.org/show_bug.cgi?id=1267075 (main patch) (part 1,2,3 and 4/ without adding tests)

Part 1: call SetBlockedRequest when CSP check failed.

Remove calling CanLoadImage. Also in the case of CSP check failed, we will call SetBlockedRequest in those cases. 

Part 2: bail out early if it's for data document.

Check the document earlier before calling into imageLoader, this could save us tons of time if in the end the document isn't allowed to load images.

Part 3: cancel imgRequestProxy if asyncOpen2 failed

The problem is if we found a cache hit, then we could go through ValidateRequestWithNewChannel to validate the cache. Then if the CSP check fail(asyncOpen2() will fail), then the imgRequestProxy will remain there, and cause the timeout. 

Part 4: add a boolean to prevent calling asynchronously

Use a boolean to prevent calling SetBlockedRequest asynchronously. Also use the same boolean to prevent some evil code reentering LoadImage.

https://bugzilla.mozilla.org/show_bug.cgi?id=1373780 (part 2 and 3)

Part 2: remove calling nsContentUtils::CanLoadImage().

Part 3: add isImgSet argument.

https://bugzilla.mozilla.org/show_bug.cgi?id=1445670

Make sure to clear out the pending request when our image gets blocked by the data document policy.

Resolves #1204

### DIFF
--- a/dom/base/nsContentPolicy.cpp
+++ b/dom/base/nsContentPolicy.cpp
@@ -22,6 +22,7 @@
 #include "nsIDOMWindow.h"
 #include "nsITabChild.h"
 #include "nsIContent.h"
+#include "nsIImageLoadingContent.h"
 #include "nsILoadContext.h"
 #include "nsCOMArray.h"
 #include "nsContentUtils.h"
@@ -145,6 +146,16 @@ nsContentPolicy::CheckPolicy(CPMethod          policyMethod,
                                          decision);
 
         if (NS_SUCCEEDED(rv) && NS_CP_REJECTED(*decision)) {
+            // If we are blocking an image, we have to let the
+            // ImageLoadingContent know that we blocked the load.
+            if (externalType == nsIContentPolicy::TYPE_IMAGE ||
+                externalType == nsIContentPolicy::TYPE_IMAGESET) {
+              nsCOMPtr<nsIImageLoadingContent> img =
+                do_QueryInterface(requestingContext);
+              if (img) {
+                img->SetBlockedRequest(*decision);
+              }
+            }
             /* policy says no, no point continuing to check */
             return NS_OK;
         }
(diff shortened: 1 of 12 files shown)


win7-7:
Opening new clean pull request to fix unnecessary spaces and one indentation I noticed in this pull request.

PR #1218, opened 2019-08-25
Add support for (later versions of) the Windows 10 SDK

### LINKED ISSUE #1217: Add support for Windows 10 SDK
Splitting this off from #1213 as requested. The patch for this should be [Bug 1356493](https://bugzilla.mozilla.org/show_bug.cgi?id=1356493).

Ascrod:
Adds support for building with the Windows 10 SDK.

This resolves #1217.

### DIFF
--- a/build/moz.configure/windows.configure
+++ b/build/moz.configure/windows.configure
@@ -361,10 +361,10 @@ set_config('LIB', lib_path)
 
 option(env='MT', nargs=1, help='Path to the Microsoft Manifest Tool')
 
-@depends_win(valid_windows_sdk_dir)
+@depends_win(valid_windows_sdk_dir, valid_ucrt_sdk_dir)
 @imports(_from='os', _import='environ')
 @imports('platform')
-def sdk_bin_path(valid_windows_sdk_dir):
+def sdk_bin_path(valid_windows_sdk_dir, valid_ucrt_sdk_dir):
     if not valid_windows_sdk_dir:
         return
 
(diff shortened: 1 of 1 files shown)


mattatobin:
I have NO idea how to test this.

Ascrod:
Ah, my bad. To test, you will need to install the Windows 10 SDK, and then run mach configure with this patch in place. I have been testing on Windows 7 with version 1809 (10.0.17763.0), which is the one that comes with VS2017, so I see this in the config log:
```
INFO: checking for Windows SDK... 
DEBUG: Creating `c:\users\ascrod\appdata\local\temp\conftest.bocbbv.c` with content:
DEBUG: | #include <winsdkver.h>
DEBUG: | WINVER_MAXVER
DEBUG: Executing: `cl.EXE -I 'C:/Program Files (x86)/Windows Kits/10/include/10.0.17763.0/um' -I 'C:/Program Files (x86)/Windows Kits/10/include/10.0.17763.0/shared' -E 'c:\users\ascrod\appdata\local\temp\conftest.bocbbv.c'`
INFO: 0x0a00 in 'C:\Program Files (x86)\Windows Kits\10\'
```
Configure should complete successfully.

Note that you can install VS2017 with the C++ workload without breaking the build system for current release if you skip (or uninstall) the Windows 10 SDK. Note also that you should be able to install just the Windows 10 SDK without Visual Studio and still use this patch.

mattatobin:
Well I can't install 2017 without it busting my environment. Can you please link the proper sdk standalone?

Ascrod:
VS2017 didn't bust my environment, but I'm only using it to build UXP applications, so I have the minimum set of C++ workload components installed and nothing else. YMMV I suppose.

Try SDK version 1809 from [here](https://developer.microsoft.com/en-us/windows/downloads/sdk-archive). I haven't done it this way myself, but I would be very surprised if this is any different than installing it through the VS2017 installer. Again, if you need to build from trunk without this patch, you *should* be able to uninstall the SDK without any adverse effects.

wolfbeast:
Tested on vs2015 with this patch, it properly detects the older SDKs, so we're good there.
```
 0:08.16 checking for Windows SDK... 0x0603 in 'C:\Program Files (x86)\Windows Kits\8.1\'
 0:08.16 checking for Universal CRT SDK... 10.0.10240.0 in 'C:\Program Files (x86)\Windows Kits\10'
```

Issue #1233, opened 2019-09-27
Grid item overflows grid column when <pre> overflows width


hawkeye116477:
Go to https://www.mediamarkt.de/de/product/_motorola-moto-g7-plus-64-gb-viva-red-dual-sim-2522327.html, scroll little and you will see the problem.

On Basilisk
![Screenshot_20190927_172536](https://user-images.githubusercontent.com/19818572/65781398-dc84d380-e14b-11e9-864d-c07ee69a34aa.png)

On Firefox
![Screenshot_20190927_172629](https://user-images.githubusercontent.com/19818572/65781491-f6beb180-e14b-11e9-85cf-0d083dfd616d.png)
![Screenshot_20190927_172645](https://user-images.githubusercontent.com/19818572/65781500-fd4d2900-e14b-11e9-86e8-9ffb15939b0e.png)

So text in **Produktbeschreibung** is too long, lacks wrapping and some information is missing on **Technische Daten**.

Related: https://bugzilla.mozilla.org/show_bug.cgi?id=1418727

wolfbeast:
Assigned to @g4jc

mattatobin:
@hawkeye116477 https://github.com/MoonchildProductions/UXP/issues/1230#issuecomment-536244756

PR #1237, opened 2019-09-28
CSS Grid Improvements

### LINKED ISSUE #1230: Back-computing percentages for intrinsic sizing causing negative percentages
Just go to https://lichess.org/editor/8/8/8/8/8/8/8/8_w_-_- and you will see the difference.

On latest Basilisk
Screenshot_20190924_132458

On Firefox 69.0
Screenshot_20190924_132628

Console doesn't show interesting info, but looking at css seems that something is not right with `grid-area`. Seems that works fine from Firefox 61.

g4jc:
Resolves #1230 and #1233


mattatobin:
First off, allow me to thank you for this it seems very complete.

However, 2 things:

1) What are the test cases?
2) Will you please next time use our commit message format because in a few months I won't have any way of tracking these commits to anything.

g4jc:
1) I backported the internal reftests and used a few of them, but I used the two test cases mentioned in the issues to verify proper working order:

* [Test 1](https://lichess.org/editor/8/8/8/8/8/8/8/8_w_-_-)

* [Test 2](https://www.mediamarkt.de/de/product/_motorola-moto-g7-plus-64-gb-viva-red-dual-sim-2522327.html

2) The reasoning for using the original commit names is to track them. I could squash the commits, but it makes it more difficult to trace them when certain bug situations pop up. (e.g. there was a bug in 1434478 part 6 which has been resolved. They had refactored and renamed parts of the file, which was totally uneeded.)

mattatobin:
Buildablity, functionality and stability are looking good thus far but a cursory code review might still be warranted.

Unstables should also be produced for this.

wolfbeast:
I can take a day out of my time off to do this.

wolfbeast:
New unstables have been built and published with this change in, for testing.

trav90:
Linux unstable has also been updated with this change in.

mattatobin:
@trav90 Yeah.. about that.. Uhh.. We need to communicate. I apparently released an unstable too today.

trav90:
@mattatobin I sent you a PM on the forum.

Issue #1234, opened 2019-09-28
libwebp NEON-specific files are missing


dimag0g:
The issue happens when building basilisk for Raspbian Buster (using gcc-4.9, but that's not a part of the issue). 3 NEON-specific files are missing in libwebp, causing the build to fail due to missing functions defined in those files. 

The issue can be reproduced by adding `ac-add-option --with-fpu=neon-vfpv4` and building with an `armhf` toolchain.

I already have a fix which consists in adding missing files from `https://github.com/webmproject/libwebp/tree/1.0.2`. Let me know if a pull request is necessary.

wolfbeast:
Go for it.

JustOff:
For completeness, [update.sh](https://github.com/MoonchildProductions/UXP/blob/master/media/libwebp/update.sh) needs to be updated as well.

dimag0g:
@JustOff Thanks for the hint. Can you tell me if I should make another pull request? AFAIK the original pull request (that's already closed) isn't easily amended.

JustOff:
Unfortunately, I noticed that a little late, so yes, you need to make a new PR.

mattatobin:
Don't feel too bad I always forget those scripts I never use but might be used at some point.

PR #1238, opened 2019-09-29
Add libwebp NEON-specific files

dimag0g:
Fix for issue #1234


mattatobin:
Why are these being always built?

wolfbeast:
Although it's not how we normally do these things, it's the way libwebp is done -- include all files, but exclude the file bodies based on config flags passed to them.

mattatobin:
Well, if that is the way webp wants it, so be it.

I'll do a quick build, then.

PR #1269, opened 2019-11-03
Expose sndio as a build option for any supporting system

### LINKED ISSUE #1265: Expose sndio as a build option for any supporting system
Our current implementation of `--enable-sndio` does not allow it to build on all supporting systems. By design it was only enabled for OpenBSD. We should make it possible to build on ALL supporting systems. It is confirmed working on GNU/Linux without any issues.

I have a patch for this already and will rebase now that #1251 is in the tree.

g4jc:
Resolves #1265


mattatobin:
Depends on libcubeb PR

PR #1281, opened 2019-11-12
Remove telemetry toolkit component

softwaretechone:
This removes the telemetry toolkit components and callers throughout the tree.

Tag #21

Copied from #1074 for clarity:
The purpose of this pull request is to remove the telemetry folder that is located in the toolkit/components folder.

I achieved this by first deleting all the files in the toolkit/components/telemetry folder. Second, of all the files in the folder, Telemetry.h was the most widely utilized file. That is why a large number of the commits mention Telemetry.h. Third, I deleted all the telemetry/health report files that were not in the toolkit/components/telemetry folder but were dependent to files within the folder. Fourth, I cleaned up any leftoever telemetry/health reporting/crash reporting code from the code base that utilized the files in the telemetry folder. Also, I deleted any telemetry/health reporting/crash reporting code that might not have used the files in the telemetry folder but was tied together with telemetry/health reporting/crash reporting code that was. Finally, the successful removal of the telemetry folder was due to removing the folder as a whole and not by increments due to how closely tied all the telemetry code was.


wolfbeast:
Updated the PR title and added a description.

toolkit/components/terminator/nsTerminator.cpp
```diff
@@ -398,14 +397,10 @@ nsTerminator::StartWatchdog()
 }
 
 // Prepare, allocate and start the writer thread. By design, it will never
-// finish, nor be deallocated. In case of error, we degrade
-// gracefully to not writing Telemetry data.
+// finish, nor be deallocated.
 void
 nsTerminator::StartWriter()
```
wolfbeast: It looks like this is only used for telemetry (it would fall back to no operation if telemetry recording isn't possible).
We can probably get rid of the entire writer component.

toolkit/components/terminator/nsTerminator.cpp
```diff
@@ -398,14 +397,10 @@ nsTerminator::StartWatchdog()
...
 {
-  if (!Telemetry::CanRecordExtended()) {
-    return;
-  }
```
wolfbeast: This should always return when there's no telemetry, effectively making this function a stub.

xpcom/build/LateWriteChecks.cpp
```diff
@@ -119,7 +117,7 @@ LateWriteObserver::Observe(IOInterposeObserver::Observation& aOb)
   }
 
   // If we have shutdown mode SCM_NOTHING or we can't record then abort
```
wolfbeast: remove "or we can't record"

xpcom/build/LateWriteChecks.cpp
```diff
@@ -165,48 +159,8 @@ LateWriteObserver::Observe(IOInterposeObserver::Observation& aOb)
 #endif
 
   SHA1Stream sha1Stream(stream);
```
wolfbeast: `stream` and `sha1` below are unused and should be deleted.

xpcom/threads/ThreadStackHelper.cpp
```diff
@@ -140,16 +139,12 @@ class ScopedSetPtr
 } // namespace
 
 void
-ThreadStackHelper::GetStack(Stack& aStack)
+ThreadStackHelper::GetStack()
```
wolfbeast: This entire function is effectively a stub, even more so if not passing back anything from it.
All it does is gather info, and then it does nothing with it because the stack filling functions `FillStackBuffer()` and `FillThreadContext()` are already stubbed out. All of that can go.

wolfbeast:
A quick try build in its current state looks good, by the way. No build bustage or functional issues seen in casual tests.

mattatobin:
Yeah and this is gonna bust my work and my applications when it lands and cause merge conflicts with mailnews-work.

This is why I specifically told you to wait AND the commit messages don't reference the issue.

wolfbeast:
No reason to not review what was submitted so far, and point out changes that need to be made. Adapting it to code changes with the mailnews landing should be fairly trivial.

mattatobin:
Except there is some telemetry in MailNews and there is corresponding telemetry in Interlink its self.

But sure.

wolfbeast:
Well let me close this PR then; @softwaretechone take the notes made to create a new PR after mailnews has been merged as explicitly requested, and it also allows you to make the commit messages follow the preferred format `Issue #21 - Commit-description-goes-here`

Issue #1279, opened 2019-11-11
Implement regex lookaround


wolfbeast:
Our irregexp has lookahead functionality but not (negative) lookbehind (special case regex).

Some websites would prefer to be able to use this, so it'd be a good enhancement to implement lookbehind in our JS regex module.
(now if only the old implementation was still used this would be trivial as it was a lot more maintainable...)

wolfbeast:
Blocks #1282

wolfbeast:
Resolved in ce0dd36a78814c59950fde6c19413c1f7ea85ee1

g4jc:
Getting build failures on trunk after this merge.

`.../js/src/builtin/TestingFunctions.cpp:3906:19: error: ‘RegExpLookahead’ is not a member of ‘js::irregexp’`

Fix appears trivial, assuming this is all that is needed to resolve this bug.

Possible patch:

```
diff --git a/js/src/builtin/TestingFunctions.cpp b/js/src/builtin/TestingFunctions.cpp
index 4363c7aed..072a72b75 100644
--- a/js/src/builtin/TestingFunctions.cpp
+++ b/js/src/builtin/TestingFunctions.cpp
@@ -3900,10 +3900,10 @@ ConvertRegExpTreeToObject(JSContext* cx, irregexp::RegExpTree* tree)
             return nullptr;
         return obj;
     }
-    if (tree->IsLookahead()) {
-        if (!StringProp(cx, obj, "type", "Lookahead"))
+    if (tree->IsLookaround()) {
+        if (!StringProp(cx, obj, "type", "Lookaround"))
             return nullptr;
-        irregexp::RegExpLookahead* t = tree->AsLookahead();
+        irregexp::RegExpLookaround* t = tree->AsLookaround();
         if (!BooleanProp(cx, obj, "is_positive", t->is_positive()))
             return nullptr;
         if (!TreeProp(cx, obj, "body", t->body()))

```

wolfbeast:
Please make a PR for this. Fix is correct.

Not sure why this didn't pop up during my development?

wolfbeast:
Oh, I see now it was already fixed through a PR. PLEASE make sure a PR to fix an issue mentions it in its description also, so it gets linked in the issue at hand, next time.

wolfbeast:
Backed out for causing #1362

PR #1290, opened 2019-11-15
Update TestingFunctions.cpp for regex lookbehind changes

g4jc:
This fixes debug builds


wolfbeast:
Resolves build bustage in #1279

Issue #1259, opened 2019-10-27
Remove Adobe Primetime Support


mattatobin:
We should exterminate the remaining support code for Adobe Primetime.

JustOff:
String changes were pushed to CrowdIn.

PR #1293, opened 2019-11-16
Adjust Widewine strings after removing Adobe Primetime support

JustOff:
Just what it says on the tin.

Tag #1259.


adeshkp:
I actually avoided this so you don't have to update langpacks.

JustOff:
In fact, these string names should have been fixed anyway; it's weird how Mozilla made such a mess there.

mattatobin:
Please do not comment on commits them selves.

PR #1297, opened 2019-11-18
Fix a compiler warning in ipc/chromium/src/base/sys_info_posix.cc.

athenian200:
Someone found a compiler warning generated by a typo in code that I wrote, so I was asked to fix it and tag issue #457. I also removed some trailing whitespace I left in the file while I was doing that.


wolfbeast:
Thanks!

Issue #1287, opened 2019-11-13
Implement asynchronous iterators (ES2018)


wolfbeast:
I'm not entirely sure if we really need this; it seems very much a corner case use scenario.

But for standards compliance, we should at least investigate it.

Blocks #1282

g4jc:
I really dislike Async Iteration because of the parser changes (it's what held back parser_tuneup branch before), but because I already have most of this logged I will attempt it again.

wolfbeast:
As said I'm not even sure we need this for anything (aside from the potential Magpie Syndrome users)

wolfbeast:
Blocks #1322

wolfbeast:
The spec is a little unclear on some points so we've discussed things to get these in while dodging some potentially dangerous corner cases that Mozilla found by fuzzing (and which are using some things that would not be used in the wild) for which the solution were refactoring and potentially opening up much worse issues (including sec)

wolfbeast:
@g4jc Were you going to open up a pull request for this or... ?

g4jc:
I was awaiting confirmation that we landed all the additional spec you would like, but yes I can submit a PR. It builds and has the implementation inspired by FF57, avoiding a few refactors.

wolfbeast:
Sorry if I wasn't entirely clear in my e-mail and the above comment in that case. We agreed the skipped bits were not actually all that relevant, and if they might become relevant in the future (although my suspicion will be that this entire feature won't see much use in the wild at all) then we can always revisit.

I was waiting on you since it blocks my intended cleanup work ;P

PR #1327, opened 2019-12-17
Implement Async Iteration in SpiderMonkey

g4jc:
Tag #1287. Please make tests on all platforms. Confirmed working on GNU/Linux.


JustOff:
Could someone please suggest me the right way to test this using new comm-central repo style?

I guess it should be something like this:
1. Create test branch in UXP repo
2. Create test branch in Pale-Moon repo
3. Edit `.gitmodules` in Pale-Moon repo to point it to test branch in UXP repo
4. Remove `platform` folder in Pale-Moon repo, then init & update submodule
5. Build in Pale-Moon repo

Am I correct and maybe there is a simpler method?

mattatobin:
You don't HAVE to use the submodule for development. So you can git submodule deinit and clone UXP directly to platform/ and manage it manually and not try and manage the submodule configuration which for you, you shouldn't be messing with.

The submodule is just very nice to use for doing builds as the commit pointer points where it should.

JustOff:
@mattatobin, thank you for the clarification.

JustOff:
I have to report that Pale Moon win32 builds just fine and passes the following [simple tests](https://www.codementor.io/@tiagolopesferreira/asynchronous-iterators-in-javascript-jl1yg8la1):

1. Async Iterators
```javascript
function asyncIterator() {
  const array = [1, 2];
  return {
    next: function() {
      if (array.length) {
        return Promise.resolve({
          value: array.shift(),
          done: false
        });
      } else {
        return Promise.resolve({
          done: true
        });
      }
    }
  };
}
```
2. for-wait-of Loop
```javascript
var iterator = asyncIterator();

(async function() {
    await iterator.next().then(console.log); // { value: 1, done: false }
    await iterator.next().then(console.log); // { value: 2, done: false }
    await iterator.next().then(console.log); // { done: true }
})();
```

3. Async Generators
```javascript
var asyncIterable = {
  [Symbol.asyncIterator]: asyncIterator
};

(async function() {
  for await (const item of asyncIterable) {
    console.log(item);
    // 1 2
  }
})();
```

wolfbeast:
r+ from me. tested for possible regressions in ES6, ES2016+ and all looks good.

PR #1335, opened 2019-12-23
Align document.open() with the overhauled specification

### LINKED ISSUE #1118: Align document.open() with overhauled spec
The new spec for this is much simpler and will be expected behavior in the not-too-distant future for all browsers. As a bonus it will remove a lot of complexity from our code.

We should follow [BMO Bug# 1489308](https://bugzilla.mozilla.org/show_bug.cgi?id=1489308) as a guideline for implementing these changes considering it's still hooked in to many things.
When implementing the overhauled spec, note should be taken of Bug 1544670 as well to prevent a potential sec issue.

wolfbeast:
This was very involved work but it seems to be working as it should now.

TL;DR this changes the behavior to align with other browsers and the spec regarding `document.open()` to replace content in the current window instead of creating a new document window and treating it as a navigation.

When landed, there still are a few follow-ups that can land to further clean up the code and remove what is no longer needed (including the whole `wyciwyg://` protocol and handler) now the handling of `document.open()/write()/close()` has been greatly simplified. These will be filed separately.

This resolves #1118


wolfbeast:
This has been on the unstable build for a number of days without any perceived/reported issues, so should be good to merge.

PR #1346, opened 2020-01-07
Issue #1345 - Implement non-standard legacy CSSStyleSheet rules

### LINKED ISSUE #1345: Implement non-standard legacy CSSStyleSheet rules
Background as researched by @JustOff 

> It all started more than twenty years ago, in October 1997, when Microsoft released Internet Explorer 4.0, where they added the addRule method to the CSS StyleSheet object (for the sake of interest, I even checked this using a Windows 95 VM box). Despite the fact that this method never became part of web standards, in May 2005, Apple added it to the WebCore library, which later became part of the Blink - the current browser engine used by Google. And, since Google today almost exclusively dictates Internet rules, so webmasters used this outdated and non-standard method in the wild, in May 2019, Mozilla was forced to do the same. 
> And, I'm afraid that in order to ensure web compatibility, we have no choice but to follow the mainstream.

A typical example of Google's "we never change existing code even if it's not standard".
Of course with mainstream aligned, this is now going to be on the standards track and the W3C has adopted this in the [CSSWG drafts](https://github.com/w3c/csswg-drafts/pull/3900). So we now want this. @JustOff has offered to pour this into a PR; assigning him.

JustOff:
Implement non-standard `CSSStyleSheet.rules`, `CSSStyleSheet.addRule` and `CSSStyleSheet.removeRule`.

Based on https://bugzilla.mozilla.org/show_bug.cgi?id=1545823, resolves #1345.

### DIFF
--- a/dom/webidl/CSSStyleSheet.webidl
+++ b/dom/webidl/CSSStyleSheet.webidl
@@ -26,4 +26,12 @@ interface CSSStyleSheet : StyleSheet {
   unsigned long insertRule(DOMString rule, optional unsigned long index = 0);
   [Throws, NeedsSubjectPrincipal]
   void deleteRule(unsigned long index);
+
+  // Non-standard WebKit things, see https://github.com/w3c/csswg-drafts/pull/3900.
+  [Throws, NeedsSubjectPrincipal, BinaryName="cssRules"]
+  readonly attribute CSSRuleList rules;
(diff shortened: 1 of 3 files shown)


wolfbeast:
Where is the removeRule implementation?

JustOff:
`rules` and `removeRule` are declared as aliases to `cssRules` and `deleteRule` in `CSSStyleSheet.webidl`.

PR #1347, opened 2020-01-08
Implement HMTL5 <dialog>

### LINKED ISSUE #1343: Implement HMTL5 <dialog>
As per title. In order to fixup our HTML5 compliance we should implement [HTML5 `<dialog>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog). Due to changes surrounding this, it blocks #252.

g4jc:
This resolves #1343. The feature is currently not enabled by default, but can be turned on by setting `dom.dialog_element.enabled` to `true`. Confirmed working via manual tests.


mattatobin:
What shade of green? Seafoam or forest green?

wolfbeast:
Definitely forest green.

Issue #1318, opened 2019-12-12
Make MOZ_PHOENIX_EXTENSIONS pref controlled


mattatobin:
In Milestone 29 we want to make the Dual-GUID System have a controlling preference will be disabled by default unless the profile already contains an active Phoenix extension.

wolfbeast:
Taking this.

wolfbeast:
Unassigning myself to focus on other things first.

wolfbeast:
With the various discussions happening around this topic it's been decided to make this an installation restriction and to change the way people are informed about compatibility mode add-ons so as to not break this for already-installed extensions. To do this properly would require a bigger rewrite of the add-on manager which should not be blocking progress at the moment.

wolfbeast:
It was decided this isn't a good approach due to technical limitations of the code and how it's all been "optimized" by Mozilla, making it impossible to properly control the various parts with a pref.

JustOff:
One way or another, I just hope all restrictions will be applied as gently as possible.

PR #1365, opened 2020-01-21
Add extensions.phoenixCompatibility pref

wolfbeast:
This changes the extension system's dual-GUID option so it is controlled
at run-time via preference `extensions.phoenixCompatibility`.
This defaults to `false` (don't allow Firefox-only extensions)

First part for #1318 -- still needs a migration path.


mattatobin:
r- because it still needs ifdef'd and has changes I don't get.

wolfbeast:
Sorry, I misunderstood your intention then about it being additional functionality.

What changes exactly don't you get?

mattatobin:
This hack is for Pale Moon only it should not be in other applications. It makes no sense and having a preference existing means it can be flipped.

wolfbeast:
Changes made. Please re-review.

Note, in case it's not clear this keeps the Firefox/native indicator hidden in the Add-on manager unless the pref is set to "true", because it wouldn't make any sense to restrict to Pale Moon extensions and still show the indicator.

wolfbeast:
As discussed in IM this causes issues with disabled extensions not being re-enable-able.
Closing this PR for now.

Issue #1366, opened 2020-01-22
Remove showModalDialog


g4jc:
As per [IRC](https://freenode.logbot.info/binaryoutcast/20200122#c3122253-c3122304) discussion, this blocks #1344. It was disabled in PM 27.6.0 and was a useless proprietary IE4 feature that nobody uses anymore and the existing implementation doesn't even work.

https://docs.microsoft.com/en-us/previous-versions/ms536759(v=vs.85)

wolfbeast:
Unfortunately building on windows is busted with this merged.

```
54:57.29 nsGlobalWindow.obj : error LNK2019: unresolved external symbol "public: static struct alrea
dy_AddRefed<class nsGlobalWindow> __cdecl nsGlobalWindow::Create(class nsGlobalWindow *)" (?Create@n
sGlobalWindow@@SA?AU?$already_AddRefed@VnsGlobalWindow@@@@PAV1@@Z) referenced in function "public: v
irtual enum nsresult __thiscall nsGlobalWindow::SetNewDocument(class nsIDocument *,class nsISupports
 *,bool)" (?SetNewDocument@nsGlobalWindow@@UAE?AW4nsresult@@PAVnsIDocument@@PAVnsISupports@@_N@Z)
54:57.29
54:57.29 nsDocShell.obj : error LNK2001: unresolved external symbol "public: static struct already_A
ddRefed<class nsGlobalWindow> __cdecl nsGlobalWindow::Create(class nsGlobalWindow *)" (?Create@nsGlo
balWindow@@SA?AU?$already_AddRefed@VnsGlobalWindow@@@@PAV1@@Z)
54:57.29
54:57.29 nsGlobalWindow.obj : error LNK2019: unresolved external symbol "protected: void __thiscall
nsGlobalWindow::InitWasOffline(void)" (?InitWasOffline@nsGlobalWindow@@IAEXXZ) referenced in functio
n "public: static struct already_AddRefed<class nsGlobalChromeWindow> __cdecl nsGlobalChromeWindow::
Create(class nsGlobalWindow *)" (?Create@nsGlobalChromeWindow@@SA?AU?$already_AddRefed@VnsGlobalChro
meWindow@@@@PAVnsGlobalWindow@@@Z)
54:57.29
54:57.29 xul.dll : fatal error LNK1120: 2 unresolved externals
```

wolfbeast:
[...]

PR #1367, opened 2020-01-22
Completely remove showModalDialog

g4jc:
This resolves Issue #1366


g4jc:
Could use some review regarding [m-c bug 1587922](https://hg.mozilla.org/mozilla-central/rev/e2e7da019485) which was used as reference for follow up removal. I didn't change `SetArguments(aArgv)` due to build bustage, but all the core functionality and tests have been removed.

Issue #1344, opened 2020-01-07
Implement Custom Elements v1


g4jc:
We have most of Custom Elements v0 in the tree, this was an evolving spec that has since moved on to v1. In order to be compatible with websites using Custom Elements v1, we need to update existing code.

https://developers.google.com/web/fundamentals/web-components/customelements

Blocks #252

wolfbeast:
Looks like we got lucky with Mozilla having something half-finished in our tree for once.
Can you indicate in which order all this is supposed to land? Does #1343 block this?

g4jc:
> Can you indicate in which order all this is supposed to land? Does #1343 block this?

Yes, while it may be possible to implement CE without <dialog>, landing #1343 first makes it much easier. Both cause changes in the HTML Tree Binding, so this will land later.

mattatobin:
Blocks #1361

g4jc:
We definitely still need to land [
customElements.define must upgrade custom element in shadow-including tree order
](https://bugzilla.mozilla.org/show_bug.cgi?id=1326028) for web  compat. Unfortunately without CSS selectors, testing will be problematic.

Issue #252, opened 2018-04-24
Consider moving getElementsByName up to document.prototype


wolfbeast:
https://dom.spec.whatwg.org/#domimplementation
https://dom.spec.whatwg.org/#interface-document
https://html.spec.whatwg.org/multipage/dom.html#the-document-object
> Let namespace be the HTML namespace, if the context object is an HTML document or context object’s content type is "application/xhtml+xml", and null otherwise. 

__An example (for Scratchpad):__

``` javascript
alert((new DOMParser).parseFromString(
"<html><body/></html>","text/html").getElementsByName);
alert((new DOMParser).parseFromString(
"<html xmlns='http://www.w3.org/1999/xhtml'><body/</html>",
"applicationxhtml+xml").getElementsByName);
```

__Actual results:__

```
function getElementsByName() {
    [native code]
}
undefined
```

__Expected results:__

```
function getElementsByName() {
    [native code]
}
function getElementsByName() {
    [native code]
}
```

Some test/personal branch/commit (with the expected result):
[_testBranch_dom_document_1](https://github.com/janekptacijarabaci/moebius/tree/_testBranch_dom_document_1)

Bug(s) - RESOLVED FIXED:
https://bugzilla.mozilla.org/show_bug.cgi?id=1415176

But see also ("META" - "Merge HTMLDocument into Document"):
https://bugzilla.mozilla.org/show_bug.cgi?id=897815

Moving body up to Document.prototype
https://bugzilla.mozilla.org/show_bug.cgi?id=1276438
(AFAIK: it does not look so easy)
Depends on (IMHO):
(EditorBase should cache pointer to mSelConWeak, mPlaceHolderTxn and mDocWeak)
https://bugzilla.mozilla.org/show_bug.cgi?id=1372829

g4jc:
Going to attempt this, a lot of dependencies so it may be awhile, but once we get DOM up to speed it should be easier to port anything else going forward.

PR #1374, opened 2020-01-26
Custom Elements v1

g4jc:
This is the majority of work completed for #1344. Due to how this lands, it is also the initial work for #252. Future follow up will included ShadowDOM changes.


g4jc:
Due to the intensive HTML Parser changes, testing should be done on all platforms.

Issue #1355, opened 2020-01-16
Improve performance of table drawing again


wolfbeast:
Since we've implemented #146 to enable relative positioning and sticky positioning on table parts, the larger display lists resulting from table drawing have caused performance regression. @win7-7 has done research what has been done at the Mozilla side to follow up on this and improve performance.
This is a tracking issue to land perf improvements against which relate to table drawing post-implementation of #146

mattatobin:
This will have to wait until #1344 has a partial landing or is completed unless it stays well out of the way of https://github.com/MoonchildProductions/UXP/compare/master...g4jc:custom_elementsv1#files_bucket which is actively being developed on at the moment.

mattatobin:
@win7-7 You have a go to do the things.

mattatobin:
@win7-7 If there is more to this it might be a good idea to step up the contribution.

mattatobin:
Is this done yet?

wolfbeast:
No, there's one more part but it's more involved.

mattatobin:
Is there an eta on the finsl part of this issue?

Are we gonna let that final part block this milestone? What is the final part?

wolfbeast:
Any further issues related to this should be their own issue. Performance of table drawing is close enough to what it was before -- some regression remains expected because of the larger display list generation with the new method required for table element positioning per-spec.

wolfbeast:
Backed out #1463 for Busting the Web™

mattatobin:
If you choose to revisit this please explain what you are going to do and preferably for a MozRefactor like this please follow an interpretation of the bug progression if that is how you are approaching this.

Additionally, this is now back in the queue and @g4jc 's `dom` work on `ShadowDOM` and `customElements` takes priority. It is ok if this doesn't make the milestone.

win7-7:
@wolfbeast you want to restore Layout\reftests\table-background\reftest.list back to original version that was in UXP tree as like said in commit:  "Layout\reftests\table-background\reftest.list should have only additions to the end of it. Revert incorrect additions." I accidentally changed middle of it while intending change only end when comparing it to Mozilla version of file during attempting to get final part/table columns change done so it should be restored to original UXP version.

mattatobin:
Could you do this?

win7-7:
[...]

PR #1376, opened 2020-01-30
 Speed up the traversal of a table row frame's child cells

win7-7:
First optimization done to  #1355.

https://bugzilla.mozilla.org/show_bug.cgi?id=1409140

GetFirstCell/GetNextCell have all this machinery to work around non-cell kids of rows, which don't actually exist.  Also, they can be inlined, since they're just a convenience wrapper around GetFirstChild/GetNextSibling with a cast tossed in.


wolfbeast:
Marginal increase in speed confirmed (160s->133s on maze test)

PR #1379, opened 2020-02-02
Hit testing in large tables has become extremely slow

win7-7:
Second optimization done to #1355.

https://bugzilla.mozilla.org/show_bug.cgi?id=1369141

This is a regression from bug 929484.

Steps to reproduce:
 1. Go to https://hg.mozilla.org/releases/mozilla-beta/annotate/cf76e00dcd6f/layout/painting/FrameLayerBuilder.cpp and wait for it to load.
 2. Move your mouse around over the code on the page.
 3. Watch how the row highlight changes.

It's very janky.

This second optimization does not change performance on Maze test but is performance improvement to regression in: https://hg.mozilla.org/releases/mozilla-beta/annotate/cf76e00dcd6f/layout/painting/FrameLayerBuilder.cpp also caused by 929484


mattatobin:
Please squash this when merged.

layout/tables/nsTableFrame.cpp
```diff
@@ -1358,7 +1358,7 @@ nsTableFrame::DisplayGenericTablePart(nsDisplayListBuilder* aBuilder,
       }
       PaintRowGroupBackgroundByColIdx(rowGroup, aFrame, aBuilder, aLists, aDirtyRect, colIdx, offset);
     }
-  } else {
+  } else if (isVisible) {
```
wolfbeast: I don't think this is correct. I just added this -- cell background visibility should depend on visibility of the cell, not the row/col(group). What did you base this on?

win7-7: https://hg.mozilla.org/mozilla-central/rev/3f243b19cba1

To Mozilla-central version of patch. https://bugzilla.mozilla.org/show_bug.cgi?id=1395312

wolfbeast: Ah, I see. Right, for background image alignment it would still have to be group aligned. Thanks for catching that!

wolfbeast:
Can you please rebase and give this PR a better title describing what it does, not what the issue problem is?

win7-7:
I will do new pull request.

PR #1388, opened 2020-02-03
Devirtualize GetRowSpan/GetColSpan

### LINKED ISSUE #1386: Devirtualize GetRowSpan/GetColSpan
These virtual functions are slow -- this should improve performance in layout on some sites.

Depends on #1384

win7-7:
https://bugzilla.mozilla.org/show_bug.cgi?id=1373095

nsTableCellFrame::GetRowSpan() is a tad slow: It's at ~1.5% on the perf log for the Netflix use case reported in bug 1373058, which seems a bit too much.

Resolves #1386

### DIFF
--- a/layout/tables/nsTableCellFrame.cpp
+++ b/layout/tables/nsTableCellFrame.cpp
@@ -740,7 +740,7 @@ nsTableCellFrame::GetColSpan()
     dom::Element* elem = mContent->AsElement();
     const nsAttrValue* attr = elem->GetParsedAttr(
       MOZ_UNLIKELY(elem->IsMathMLElement()) ? nsGkAtoms::columnspan_
-                                              : nsGkAtoms::colspan);
+                                            : nsGkAtoms::colspan);
     // Note that we don't need to check the tag name, because only table cells
     // (including MathML <mtd>) and table headers parse the "colspan" attribute
     // into an integer.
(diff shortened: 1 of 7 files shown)


wolfbeast:
... why are we looking at changing things for something only taking 1.5% of page layout time?

layout/tables/nsTableCellFrame.cpp
```diff
@@ -711,33 +711,39 @@ nsTableCellFrame::GetCellBaseline() const
...
-  nsGenericHTMLElement *hc = nsGenericHTMLElement::FromContent(mContent);
 
   // Don't look at the content's colspan if we're a pseudo cell
-  if (hc && !StyleContext()->GetPseudo()) {
-    const nsAttrValue* attr = hc->GetParsedAttr(nsGkAtoms::colspan);
+  if (!StyleContext()->GetPseudo()) {
+    dom::Element* elem = mContent->AsElement();
+    const nsAttrValue* attr = elem->GetParsedAttr(
+      MOZ_UNLIKELY(elem->IsMathMLElement()) ? nsGkAtoms::columnspan_
+                                              : nsGkAtoms::colspan);
```
wolfbeast: Nit: whitespace

mattatobin:
Is there a test case?

win7-7:
No. Since Netflix has rewritten the DVD Queue page that this patch improves there is no currently way to test this (and Netflix account would be required for somebody to test it anyway) but it should still be small performance improvement.

wolfbeast:
So, I ask again, why are we even looking at changing this code if it's on the verge of beng irrelevant? is this a prerequisite for a later patch or...? some way to test functionality would also be good, is there no synthetic tests we can pull this through to prevent breakage?

PR #1393, opened 2020-02-06
 Do less work for columns not in the desired set in PaintRowGroupBackgroundByColIdx

win7-7:
Third optimization done to #1355.

Right now we only check the column after checking the cell visibility, but the cell visibility check is probably more expensive.

Also, we can stop the iteration once our index is larger than the largest index in the array.

This is based on https://bugzilla.mozilla.org/show_bug.cgi?id=1409154

Also includes crash fix/ https://bugzilla.mozilla.org/show_bug.cgi?id=1409787

This is about 15 tiles per second improvement to Maze test. I tested this with New Moon build.


wolfbeast:
Tested with several functionality references, looks good. performance improvement confirmed.

PR #1458, opened 2020-02-20
Fix buildability of builds without devtools-server

### LINKED ISSUE #316: The codebase won't build without DevTools Server because of heapsnapshot
Seems DOM expects `devtools/shared/heapsnapshot` to be there. I know it is absolutely required for DevTools them selves but I need to investigate anything else that might expect it and if nothing.. Make the DOM code that uses it conditional.

- [x] Figure out what the hell `heapsnapshot` actually is and if we want to keep it
- [x] Always build `heapsnapshot` regardless of the rest of the devtools for now
- [x] Make the DOM bits conditional
- [x] Move headsnapshot into DOM

wolfbeast:
This resolves #316

### DIFF
--- a/dom/performance/Performance.h
+++ b/dom/performance/Performance.h
@@ -96,8 +96,10 @@ public:
 
   IMPL_EVENT_HANDLER(resourcetimingbufferfull)
 
+#ifdef MOZ_DEVTOOLS_SERVER
   virtual void GetMozMemory(JSContext *aCx,
                             JS::MutableHandle<JSObject*> aObj) = 0;
+#endif
 
   virtual nsDOMNavigationTiming* GetDOMTiming() const = 0;
 
--- a/dom/performance/PerformanceMainThread.cpp
+++ b/dom/performance/PerformanceMainThread.cpp
@@ -11,6 +11,7 @@
 namespace mozilla {
 namespace dom {
 
+#ifdef MOZ_DEVTOOLS_SERVER
 NS_IMPL_CYCLE_COLLECTION_CLASS(PerformanceMainThread)
 
 NS_IMPL_CYCLE_COLLECTION_UNLINK_BEGIN_INHERITED(PerformanceMainThread,
@@ -36,6 +37,10 @@ NS_IMPL_CYCLE_COLLECTION_TRACE_END
 
 NS_IMPL_ADDREF_INHERITED(PerformanceMainThread, Performance)
 NS_IMPL_RELEASE_INHERITED(PerformanceMainThread, Performance)
+#else
+NS_IMPL_CYCLE_COLLECTION_INHERITED(PerformanceMainThread, Performance
+                                   mTiming, mNavigation, mDocEntry)
+#endif
 
 // QueryInterface implementation for PerformanceMainThread
 NS_INTERFACE_MAP_BEGIN_CYCLE_COLLECTION(PerformanceMainThread)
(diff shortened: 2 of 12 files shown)


JustOff:
It seems something missing prevents linking Pale Moon:
```
Unified_cpp_dom_performance0.obj : error LNK2019: unresolved external symbol "class JSObject * __cdecl js::gc::NewMemoryInfoObject(struct JSContext *)" (?NewMemoryInfoObject@gc@js@@YAPAVJSObject@@PAUJSContext@@@Z) referenced in function "public: virtual void __thiscall mozilla::dom::PerformanceMainThread::GetMozMemory(struct JSContext *,class JS::MutableHandle<class JSObject *>)" (?GetMozMemory@PerformanceMainThread@dom@mozilla@@UAEXPAUJSContext@@V?$MutableHandle@PAVJSObject@@@JS@@@Z)
xul.dll : fatal error LNK1120: 1 unresolved externals
```

wolfbeast:
providing more output would help.

mattatobin:
Did you clobber? Cause I have built three times since this landed and everything is a-ok on my end.

wolfbeast:
`PerformanceMainThread::GetMozMemory` is `#ifdef`ed out so it shouldn't complain. You probably need to clobber if you swap that mozconfig option.

JustOff:
> providing more output would help.

Unfortunately nothing else meaningful was not fit in the console, see https://ctxt.io/2/AABAFcVUFA

> Did you clobber?

I did not, but `configure` was forced by some of the latest changes. I'll try clobber, thanks.

JustOff:
Unfortunately, clobber did not help ...

wolfbeast:
Well, I tried looking at your output but "Paste expired or not found. Create your own Context here."  -- not very useful, is it?

JustOff:
Sorry, it has expired somehow ... here is the latest: https://justpaste.it/33c8s

wolfbeast:
Thanks. makes no sense. the function and symbol reference point to `PerformanceMainThread::GetMozMemory(JSContext *aCx, JS::MutableHandle<JSObject*> aObj)` in `/dom/performance/PerformanceMainThread.cpp` and that has the conditionals around it to exclude it when building without devtools-server, so this link error makes no sense... :confused: 
I have also built 2 times already with clobber since that commit and it passes without complaints.

JustOff:
And here is `.mozbuild\last_log.json`: [last_log.json.zip](https://github.com/MoonchildProductions/UXP/files/4233070/last_log.json.zip)

wolfbeast:
> And here is .mozbuild\last_log.json: last_log.json.zip

Wait... you're telling me it fails with `devtools-server` enabled?

JustOff:
> that has the conditionals around it to exclude it when building without devtools-server

But by default it's not excluded, isn't it?

```
D:\UXP-build-palemoon
λ grep MOZ_DEVTOOLS_SERVER config.status
    'MOZ_DEVTOOLS_SERVER': '1',
    'MOZ_DEVTOOLS_SERVER': '1',
```

> Wait... you're telling me it fails with `devtools-server` enabled?

yes

wolfbeast:
Problem was the define wasn't passed down to js and parts of this code are in js -- follow-up incoming.

JustOff:
Thanks, everything is fine after https://github.com/MoonchildProductions/UXP/commit/5e21bf1d544f8d15e73cf7cb4ca793addd4bc38d.

Issue #146, opened 2018-04-13
Change the way HTML tables are rendered


wolfbeast:
If we want to be able to support sticky positions on table elements, we need to change how HTML tables are rendered.

Specifically, this will need the following ported:
- Bug 929484 - Make HTML table frames (including row/cell/etc. frames) create their own display list items
- Bug 975644 - Make position: sticky work on table parts

PR #1463, opened 2020-02-25
Better way to create display items for column backgrounds

### LINKED ISSUE #1355: Improve performance of table drawing again

win7-7:
https://bugzilla.mozilla.org/show_bug.cgi?id=1409114

Resolves #1355

To get this work with UXP codebase certain amount of refactoring was necessary so Moonchild should review this carefully since Mozilla has changed lot of display list and layout code before 1409114 and getting some/necessary part of these changes for 1409114  into UXP should be carefully checked that they work as intended.

https://bugzilla.mozilla.org/show_bug.cgi?id=1388161 is necessary as it makes applying 1409114 easier without requiring aDirtyRect/more changes to get the column backgrounds improvements to work.

https://bugzilla.mozilla.org/show_bug.cgi?id=1367747 this is necessary as it adds visibility checks that again makes 1409114 easier to apply to UXP.

Part 3 can't be applied to UXP because it seems it depends on Stylo/ArcSlice: https://bugzilla.mozilla.org/show_bug.cgi?id=1550554

Part 10: Depends on one part of retained display list, does work by applying Bug 1404181 - Part 2: Track a 'visible' rect for display list building as well as the 'dirty' rect, so that we can do partial building by specifying a smaller 'dirty' rect.

Part 11:

Depends on AutoBuildingDisplayList changes/need code changes to display list.

This is about 36-38 tiles per second improvement to Maze test. Now this is much closer to performance before issue #146. Maze test is still about 22-24 tiles less than -GL build before #146 was. It might be improved by optimizing layout and display list more than currently but those should be separate issues later since main regression is resolved after this and improvements to layout or display list need research. I tested this with New Moon build with Whole Program Optimization/-GL ( I always use -GL when building New Moon).

Little off-topic:

Whole Program Optimization builds have been stable and results in significant performance gains in New Moon. Moonchild maybe you could try use -GL in unstable channel?

Also could you if you have time port following performance improvements: https://bugzilla.mozilla.org/show_bug.cgi?id=1343596 and https://bugzilla.mozilla.org/show_bug.cgi?id=1342009 as you could ask Mozilla access to security issues these improvements cause as security fixes also would be necessary to be ported along with improvements.

### DIFF
--- a/layout/base/nsDisplayList.cpp
+++ b/layout/base/nsDisplayList.cpp
@@ -721,7 +721,6 @@ nsDisplayListBuilder::nsDisplayListBuilder(nsIFrame* aReferenceFrame,
     : mReferenceFrame(aReferenceFrame),
       mIgnoreScrollFrame(nullptr),
       mLayerEventRegions(nullptr),
-      mCurrentTableItem(nullptr),
       mCurrentFrame(aReferenceFrame),
       mCurrentReferenceFrame(aReferenceFrame),
       mCurrentAGR(&mRootAGR),
@@ -731,6 +730,7 @@ nsDisplayListBuilder::nsDisplayListBuilder(nsIFrame* aReferenceFrame,
       mGlassDisplayItem(nullptr),
       mScrollInfoItemsForHoisting(nullptr),
       mMode(aMode),
+      mTableBackgroundSet(nullptr),
       mCurrentScrollParentId(FrameMetrics::NULL_SCROLL_ID),
       mCurrentScrollbarTarget(FrameMetrics::NULL_SCROLL_ID),
       mCurrentScrollbarFlags(0),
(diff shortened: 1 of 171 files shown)


mattatobin:
/GL may be stable but there is almost no tangible benefit and it literally doubles build time largely due to linking so.. not gonna happen.

Go faster at any costs is not goal.

wolfbeast:
Okay. Please for the future if you find that you want to rearchitecture/refactor things to apply porting of bugs, you HAVE to communicate this before you put in the work so we can see if it's something that should be done or not. At any point in time multiple people are working on code and refactoring can seriously upset their work/requiring rebases/clobbers what they've done.
Thankfully this time it seems that the impact isn't terrible because we happen to not be very busy in layout right now, but we did have to make sure that stays this way now until I've reviewed the changes and landed it.

wolfbeast:
Unfortunately this broke the web for users with display items disappearing.
See https://forum.palemoon.org/viewtopic.php?f=5&t=23904

Issue #1471, opened 2020-03-05
Pale Moon cannot be built on SPARC64 Linux


JMadgwick:
Attempting to build Pale Moon on Linux with SPARC64 architecture results in build errors. These are caused by incorrect compiler definitions. These problems can be fixed using patches from [bugzilla bug #1275204](https://bugzilla.mozilla.org/show_bug.cgi?id=1275204).
Pale Moon is then able to build on SPARC64, [see Pale Moon Forum](https://forum.palemoon.org/viewtopic.php?f=5&t=23934).

mattatobin:
So is this fixed now?

JMadgwick:
Yes, the pull request has been merged so this is no longer a problem. Apologies for not closing this already.

mattatobin:
Please don't close issues yourself also thanks!

PR #1472, opened 2020-03-05
Fix building on sparc64 Linux

JMadgwick:
Correct various pre-processor defines for sparc64 and in mozjemalloc use the JS arm64 allocator on Linux/sparc64.
This corrects build problems on Linux sparc64 and is in line with bugzilla bug #1275204.

### DIFF
--- a/toolkit/components/protobuf/src/google/protobuf/stubs/platform_macros.h
+++ b/toolkit/components/protobuf/src/google/protobuf/stubs/platform_macros.h
@@ -67,7 +67,7 @@
 #define GOOGLE_PROTOBUF_ARCH_32_BIT 1
 #elif defined(sparc)
 #define GOOGLE_PROTOBUF_ARCH_SPARC 1
-#ifdef SOLARIS_64BIT_ENABLED
+#ifdef __arch64__
 #define GOOGLE_PROTOBUF_ARCH_64_BIT 1
 #else
 #define GOOGLE_PROTOBUF_ARCH_32_BIT 1
--- a/xpcom/reflect/xptcall/md/unix/moz.build
+++ b/xpcom/reflect/xptcall/md/unix/moz.build
@@ -177,7 +177,7 @@ if CONFIG['OS_ARCH'] == 'OpenBSD' and CONFIG['OS_TEST'] == 'powerpc':
         'xptcstubs_ppc_openbsd.cpp',
     ]
 
-if CONFIG['OS_ARCH'] == 'Linux' and 'sparc' in CONFIG['OS_TEST']:
+if CONFIG['OS_ARCH'] == 'Linux' and CONFIG['OS_TEST'] == 'sparc':
     SOURCES += [
         'xptcinvoke_asm_sparc_linux_GCC3.s',
         'xptcinvoke_sparc_solaris.cpp',
@@ -205,7 +205,7 @@ if CONFIG['OS_ARCH'] == 'OpenBSD' and CONFIG['OS_TEST'] == 'sparc':
 # files for 64-bit SPARC with no ill effects, so basically the entire mess that
 # was there before is no longer needed.
 
-if CONFIG['OS_ARCH'] in ('OpenBSD', 'FreeBSD', 'SunOS') and CONFIG['OS_TEST'] == 'sparc64':
+if CONFIG['OS_ARCH'] in ('OpenBSD', 'FreeBSD', 'SunOS', 'Linux') and CONFIG['OS_TEST'] == 'sparc64':
     SOURCES += [
         'xptcinvoke_asm_sparc64_openbsd.s',
         'xptcinvoke_sparc64_openbsd.cpp',
--- a/ipc/chromium/src/build/build_config.h
+++ b/ipc/chromium/src/build/build_config.h
@@ -85,7 +85,7 @@
 #elif defined(__ppc__) || defined(__powerpc__)
 #define ARCH_CPU_PPC 1
 #define ARCH_CPU_32_BITS 1
-#elif defined(__sparc64__)
+#elif defined(__sparc__) && defined(__arch64__)
 #define ARCH_CPU_SPARC 1
 #define ARCH_CPU_64_BITS 1
 #elif defined(__sparc__)
--- a/js/src/gc/Memory.cpp
+++ b/js/src/gc/Memory.cpp
@@ -423,7 +423,7 @@ MapMemoryAt(void* desired, size_t length, int prot = PROT_READ | PROT_WRITE,
 // possibly be correct on AMD64, but for Solaris/illumos it is.	
 	
 {
-#if defined(__ia64__) || (defined(__sparc64__) && defined(__NetBSD__)) || defined(__aarch64__) || (defined(__sun) && defined(__x86_64__))
+#if defined(__ia64__) || defined(__aarch64__) || (defined(__sun) && defined(__x86_64__)) || (defined(__sparc__) && defined(__arch64__) && (defined(__NetBSD__) || defined(__linux__)))
     MOZ_ASSERT((0xffff800000000000ULL & (uintptr_t(desired) + length - 1)) == 0);
 #endif
     void* region = mmap(desired, length, prot, flags, fd, offset);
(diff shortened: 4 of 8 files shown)


JMadgwick:
This is the work required for #1471.

wolfbeast:
@athenian200 maybe you can have a look here, being the solaris guy?

mattatobin:
@JMadgwick Could you please modify your commit message to add a dash between Issue #xxx and the rest then force push your branch up?

athenian200:
Yeah, I'm willing to test it, but there is a bit of a problem with this patch.

As it stands, this patch would prevent Solaris/illumos from working on x86_64. It undoes something that I explicitly had to do in order to get x86_64 working with Solaris in the process of adding SPARC64.

Specifically, it removes (defined(__sun) && defined(__x86_64__)) from a few areas in Memory.cpp in the process of adding SPARC64 support, and Solaris won't work on x86_64 without them. The reason is very counter intuitive and has something to do with the way Solaris/illumos handles memory segmentation model on x86_64.

JMadgwick:
> Specifically, it removes (defined(__sun) && defined(**x86_64**)) from a few areas in Memory.cpp in the process of adding SPARC64 support, and Solaris won't work on x86_64 without them.
Sorry, I didn't realise that the code was different to mozilla's when applying the patches.
I will correct that when I rename the commit.

mattatobin:
I don't see where you corrected anything.. Didn't he say that you need to retain specifically his changes as to not break `SunOS` in general?

JMadgwick:
Sorry I force pushed twice as I forgot to commit the first time... The changes are in Memory.cpp, where @athenian200 had made changes which I hadn't noticed when I applied the mozilla patches. His changes are now back in place.

mattatobin:
I am not at all sure about these changes they seem to explicitly favor linux for a lot of sparc.. I also don't like that you split the preprocessor directives up onto other lines.

xpcom/reflect/xptcall/md/unix/moz.build
```diff
@@ -177,7 +177,7 @@ if CONFIG['OS_ARCH'] == 'OpenBSD' and CONFIG['OS_TEST'] == 'powerpc':
         'xptcstubs_ppc_openbsd.cpp',
     ]
 
-if CONFIG['OS_ARCH'] == 'Linux' and 'sparc' in CONFIG['OS_TEST']:
+if CONFIG['OS_ARCH'] == 'Linux' and CONFIG['OS_TEST'] == 'sparc':
```
mattatobin: You know if you keep this it will ONLY be true if linux and sparc NOT sparc and sparc64 right?

toolkit/components/protobuf/src/google/protobuf/stubs/platform_macros.h
```diff
@@ -67,7 +67,7 @@
 #define GOOGLE_PROTOBUF_ARCH_32_BIT 1
 #elif defined(sparc)
 #define GOOGLE_PROTOBUF_ARCH_SPARC 1
-#ifdef SOLARIS_64BIT_ENABLED
+#if defined(__arch64__)
```
mattatobin: This should still be #ifdef

xpcom/reflect/xptcall/md/unix/moz.build
JMadgwick: Yes, it comes from this patch. It looks like it was done more to tidy up the style (to not use "in"). It isn't strictly required for the patch.

toolkit/components/protobuf/src/google/protobuf/stubs/platform_macros.h
JMadgwick: It should. I'll correct this.

JMadgwick:
From the discussion on the bugzilla bug the reason that Linux is checked for specifically is because of differences in the memory model used in Solaris SPARC. So the `&& defined(__linux__)` is needed to prevent the Solaris SPARC build from breaking.
That said, I have no idea if Pale Moon is currently able to build on Solaris SPARC. But if it is these changes will not impact that because of the check for Linux in these changes.
Regarding the split directives, I took this from [the style of the patches](https://hg.mozilla.org/mozilla-central/diff/0cb0fe7e92f695be45e21c95907bbc6a12664f02/js/src/gc/Memory.cpp). It would be rather long on one line but it could be done with two if you would prefer.

xpcom/reflect/xptcall/md/unix/moz.build
mattatobin: Yeah but OS_TEST == 'sparc' will ONLY be a true condition IF OS_TEST is exactly 'sparc' if OS_TEST happens to be 'sparc64' for whatever reason it will NOT be true and the entire conditional will be false. Is this the intended logic?

mattatobin:
We don't follow Mozilla's current Google styling. You follow the predominate style of the file you are working in. As far as I know there are no split preprocessor directives and I am not even sure if the Python C-Preprocessor in our tree supports that.

xpcom/reflect/xptcall/md/unix/moz.build
JMadgwick: Yes, the code it's protecting is specific to sparc and not sparc64 so it should be true only for sparc (32-bit) and not sparc64. The conditional for sparc64 is at line 208, it pulls in similar files but for sparc64. Why that conditional checks for certain OS I don't know. It seems that the OS check at 208 is redundant, as if it wasn't originally required on Linux and so was excluded, but now it's included it might as well not check the OS. I didn't change this as I didn't want to diverge from the mozilla patches and make my own interpretation of it, although it looks like maybe I should have done that.

JMadgwick:
> As far as I know there are no split preprocessor directives and I am not even sure if the Python C-Preprocessor in our tree supports that.

I'll remove the line splits (and correct that `ifdef`) when I force push again. I'll wait to hear back from @athenian200 before I do that though.

xpcom/reflect/xptcall/md/unix/moz.build
mattatobin: Alright as long as you are sure about it. Python syntax is a bit odd sometimes is all.

athenian200:
You added back the part I needed to be in there, so it looks okay on that front. As for the other changes, I would still have to test them. I've never really installed Linux on SPARC before, so I will have to get that setup for a testing environment. Hopefully it doesn't take too long for me to figure it all out.

mattatobin:
So go head and make the final changes to your patch @JMadgwick

JMadgwick:
> I've never really installed Linux on SPARC before, so I will have to get that setup for a testing environment. Hopefully it doesn't take too long for me to figure it all out.

I used the [Debian sparc64 port](https://wiki.debian.org/Sparc64). The only trouble I had was getting the package signing key to work properly. I fully intend to also install Solaris but I haven't found the time yet.

While I'm at it I should also check the build on the BSDs. After @mattatobin's comment earlier I'm suspicious of the [conditional here](https://github.com/MoonchildProductions/UXP/commit/f7c9359b2ac431ac99fb4e8a1ebd799e6e8fb21c#diff-4ce2d833c7fede3dbfbdf9358f738c14R208), I suspect it can be simplified. After my changes it's pretty much just looking at a list containing all the  OS capable of running on sparc64 (except NetBSD) and then checking if the arch is sparc64. The strange thing is that if this conditional failed then no other sparc64 code is pulled in. Without it I can't see how NetBSD would be able to build? The only way to prove this would be to test it and I don't know if any BSD is able to build Pale Moon currently, a quick look at forum and it doesn't look like support is in good shape at the moment, something for another day...

mattatobin:
Ignore BSD until such time as `Hyperbola` gets theirs going. Changes are gonna have to be made anyway to accommodate it.

athenian200:
Okay, I'm kind of running into some problems with this... I can't get a straight answer online about how to install the Debian SPARC64 port. The documentation is scant, and seems to be targeted at people who already know SPARC inside out. I burned a Debian SPARC64 ISO to a CD-ROM, but I can't get it to boot up. 

Can you provide a step-by-step guide to installing Debian on a T4-1 system, for someone that's basically clueless about SPARC and only just learned about how to use a serial console cable for the first time in their lives? I'm in way over my head here.

I relied heavily on Oracle's documentation to get Solaris installed on here, and there's no similarly extensive documentation I can find about installing Linux on SPARC. From what I can gather, it seems to be basically a hack, like installing Linux on a rooted Android phone, and not some kind of normal or reasonable procedure.

JMadgwick:
I didn't have a particular problem getting the installer to boot. I simply booted it in the same way as a Solaris installer. It's probably better to use a usb, unfortunately this is more difficult as one must list all volumes first, to find the device name in order to use the boot command.

As this is off topic for this pull request it's probably better to ask a question on stack exchange or similar detailing the problem.
________________________________

athenian200:
So yeah, it turns out I can't really verify anything because I've managed to get my SPARC server into a state where it won't boot at all. It happened after a recent power outage. 

I'm not happy about it, but it looks like I'm going to have to abandon Solaris SPARC as a target platform because I no longer have the capacity to maintain it, and will just have to let you maintain the Linux SPARC build as you see fit, without any testing or knowing whether it breaks Solaris SPARC or not. 

I'm not seeing any problems on Solaris x86_64 anymore, though, so that's good news at least.

Issue #1469, opened 2020-03-03
PR_GetLibraryFilePathname is returning absolute paths in MacOS Catalina.


dbsoft:
Local builds fail on MacOS Catalina.

Steps to reproduce: ./mach configure && ./mach build && ./mach run
Result: Assertion failure: NS_IsMainThread(), at /Users/..../work/github/UXP/security/manager/ssl/nsNSSComponent.cpp:258
Expected Result: No error
Affected Versions: Likely all versions
Affected Environment: macOS 10.15 Catalina

 Apparently in 10.15 dladdr() returns the absolute path of the file instead of the path of the symlink. The application  attempts to load libsoftokn3.dylib by using PR_GetLibraryFilePathname() which uses dladdr() to determine the path to libnss3.dylib. Prior to 10.15 this would return the New Moon.app/Contents/MacOS/ path on 10.15 however it uses the real path which is /Users/..../work/github/UXP/obj-x86_64-apple-darwin19.3.0/config/external/nss/. And libsoftokn3.dylib is not located there causing NSS to fail. Mozilla fixed this by copying the files to the .app/Contents/MacOS instead of using symbolic links.

Here is the patch that fixes Mozilla builds:

https://hg.mozilla.org/releases/comm-esr60/rev/2de29b2f2c3ed933481691812f51df071a78d512

wolfbeast:
This targets application code, and should be opened in the various application repos.

mattatobin:
I am reopening this as it does affect all applications that can build with mac support. So I can reference. I will do the patches BUT I need a mac builder to provide me with a build that targets our defacto minimum of 10.8 or at least 10.9.

dbsoft:
> 
> 
> I am reopening this as it does affect all applications that can build with mac support. So I can reference. I will do the patches BUT I need a mac builder to provide me with a build that targets our defacto minimum of 10.8 or at least 10.9 with the changes in the invalid PR before I do the things.

The wiki says it should target 10.7, and that is what I am using as target for my builds.  But these are not code changes, just changing how the application bundle is created, copying the dynamic libraries instead of creating symbolic links to them.

mattatobin:
And I like to have verification and testing. How the bundle is created or how dist/bin is dealt with is just as important as tangible code changes to cpp or js.

dbsoft:
> And I like to have verification and testing. How the bundle is created or how dist/bin is dealt with is just as important as tangible code changes to cpp or js.

[...]

PR #1475, opened 2020-03-09
Catalina build fixes

dbsoft:
These patches should resolve issue #1469


wolfbeast:
if this targets application code, then it doesn't belong in this repo.

Also, it's called "Pale Moon"

dbsoft:
> r- because it targets release branches, there is no indication if these are backwards compatible, and nothing is offered to the application repositories and has invalid commit messages.

I don't know where it is supposed to go, the master branch does not have these source trees.

Also what invalid commit messages?

Issue #1053, opened 2019-04-23
Drop support for Android and drop Fennec


wolfbeast:
There are no plans whatsoever to build a Fennec client from UXP or any other app for that matter that targets Android. Any potential client would also only be interesting for users if it was a client that could sync in a Weave fashion, which is currently also non-existent in the `/mobile` client in the tree as it was ripped out in favor of FxA.

Doing this will allow us to greatly simplify code throughout the tree; any future Android client should be built on native Android engines for the moving target of Google "Standards" anyway to be useful, and would best be built from the ground up, and be a native Android developed app.

- [ ] Remove Fennec. Sorry little guy, but you have to find your own home.
- [ ] Remove Android support in our renderer/widgets
- [ ] Remove Android support in our layout engine
- [ ] Remove Android support in XPCOM
- [ ] Remove Android support in the rest of our tree
- [ ] Remove Android support in the build system

mattatobin:
Just in case anyone asks.. This shall not remove ARM architecture support.

JustOff:
There are some leftovers in `UserAgentUpdates.jsm` that were missed during [removal of android support from netwerk](https://github.com/MoonchildProductions/UXP/commit/bdad2357a1f0d2aca7f1106f3195d1fd929dc1d1#diff-888228427807bd9ad3e953ef1df43f43). I noticed this while working on improving SSUAO, and now I'm going to submit a PR to remove them.

mattatobin:
Such as?

JustOff:
These are `function readChannel(url)` and lazy module getters for `NetUtil.jsm` and `Promise.jsm`, see https://github.com/mozilla/gecko-dev/commit/78eb87174f28b941e2fd6c098e9e4d50c206e377.

mattatobin:
Oh ok.. Go for it then.

PR #1479, opened 2020-03-10
Remove no longer used function and modules from UserAgentUpdates.jsm

JustOff:
This removes no longer used `readChannel(url)` and lazy module getters for `NetUtil.jsm` and `Promise.jsm` that are left after removal of android support from netwerk.

Tag #1053.


wolfbeast:
Thanks, looks like all callers for this function were removed so it could indeed be taken out.

PR #1484, opened 2020-03-12
Give dynamic SSUAO rules priority over static ones

### LINKED ISSUE #1482: Dynamic SSUAO rules have insufficient priority
As I mentioned [here](https://github.com/MoonchildProductions/UXP/issues/1473#issuecomment-595888333), сurrently all static SSUAO rules always take precedence over those received dynamically, which does not allow using dynamic updating to adjust builtin static rules, but only to add new ones. Considering that we [are not going to](https://github.com/MoonchildProductions/Pale-Moon/issues/1733#issuecomment-595273585) do everything with dynamic SSUAO and to refuse of using pre-set overrides, but to use this service mainly for hot-fixes, this behavior looks like a serious flaw.

I propose to resolve this by giving dynamic SSUAO rules priority over static rules from the default preference branch (where pre-set SSUAO rules live), but keeping the highest priority for static rules created by the user.

JustOff:
This gives dynamic SSUAO rules priority over static, while keeps the highest one for rules created by the user.

Resolves #1482.

### DIFF
--- a/netwerk/protocol/http/UserAgentOverrides.jsm
+++ b/netwerk/protocol/http/UserAgentOverrides.jsm
@@ -30,6 +30,7 @@ XPCOMUtils.defineLazyServiceGetter(this, "ppmm",
 var gPrefBranch;
 var gOverrides = new Map;
 var gUpdatedOverrides;
+var gOldDynOverrides;
 var gOverrideForHostCache = new Map;
 var gInitialized = false;
 var gOverrideFunctions = [
@@ -55,7 +56,10 @@ this.UserAgentOverrides = {
     }
 
     UserAgentUpdates.init(function(overrides) {
-      gOverrideForHostCache.clear();
+      if (overrides == gOldDynOverrides) {
+        return;
+      }
+      gOldDynOverrides = overrides;
       if (overrides) {
         for (let domain in overrides) {
           overrides[domain] = getUserAgentFromOverride(overrides[domain]);
@@ -63,6 +67,7 @@ this.UserAgentOverrides = {
         overrides.get = function(key) { return this[key]; };
       }
       gUpdatedOverrides = overrides;
+      buildOverrides();
     });
 
     buildOverrides();
@@ -159,15 +164,22 @@ function buildOverrides() {
   if (!Services.prefs.getBoolPref(PREF_OVERRIDES_ENABLED))
     return;
 
-  let builtUAs = new Map;
   let domains = gPrefBranch.getChildList("");
 
+  // Since the static override map has the highest priority, we build it so
+  // that it includes only domains with user-set overrides and domains with
+  // default pre-set overrides that are not overridden by dynamic updates.
   for (let domain of domains) {
-    let override = gPrefBranch.getCharPref(domain);
-    let userAgent = getUserAgentFromOverride(override);
+    if (!(gUpdatedOverrides && gUpdatedOverrides.get(domain)) ||
+        gPrefBranch.prefHasUserValue(domain)) {
+      // Here we selected domains that are not dynamically overridden
+      // or have a user-set override.
+      let override = gPrefBranch.getCharPref(domain);
+      let userAgent = getUserAgentFromOverride(override);
 
-    if (userAgent != DEFAULT_UA) {
-      gOverrides.set(domain, userAgent);
+      if (userAgent != DEFAULT_UA) {
+        gOverrides.set(domain, userAgent);
+      }
     }
   }
 }


netwerk/protocol/http/UserAgentOverrides.jsm
```diff
@@ -164,10 +169,15 @@ function buildOverrides() {
...
-      gOverrides.set(domain, userAgent);
+    if (gUpdatedOverrides === null ||
+        gUpdatedOverrides === undefined ||
+        gUpdatedOverrides.get(domain) === undefined ||
+        gPrefBranch.prefHasUserValue(domain)) {
+      let userAgent = getUserAgentFromOverride(override);
+
+      if (userAgent != DEFAULT_UA) {
+        gOverrides.set(domain, userAgent);
+      }
```
wolfbeast: I think this is wrong. You'd only be setting the override if there is a user-set override or no dynamic override for the domain, ignoring _default_ pre-set overrides.
Also, the condition is overly elaborate. `!gUpdatedOverrides` should account for `null` and `undefined` (and also covering the desired `""` case). same for `!gUpdatedOverrides.get(domain)`.

netwerk/protocol/http/UserAgentOverrides.jsm
```diff
@@ -55,14 +56,18 @@ this.UserAgentOverrides = {
     }
 
     UserAgentUpdates.init(function(overrides) {
-      gOverrideForHostCache.clear();
```
wolfbeast: Why was this removed? Please explain your logic.

netwerk/protocol/http/UserAgentOverrides.jsm
```diff
@@ -55,14 +56,18 @@ this.UserAgentOverrides = {
...
+      }
+      gOldDynOverrides = overrides;
       if (overrides) {
         for (let domain in overrides) {
           overrides[domain] = getUserAgentFromOverride(overrides[domain]);
         }
         overrides.get = function(key) { return this[key]; };
       }
       gUpdatedOverrides = overrides;
+      buildOverrides();
```
wolfbeast: Why is this necessary? Please explain your logic.

netwerk/protocol/http/UserAgentOverrides.jsm
JustOff: I'm sorry, but I think you did not quite understand this condition. `gOverrides` contains rules that are checked and applied [first](https://github.com/MoonchildProductions/UXP/blob/3c8bc3da7cd537680288e8b870c21e1ae789a541/netwerk/protocol/http/UserAgentOverrides.jsm#L104-L105), before `gUpdatedOverrides`, so I put all user-set rules there, as well as default pre-set rules that are not overridden by dynamic ones. I tested it in all possible combinations, and it gives the desired result.

As for complex expressions, I just wanted to make the checks explicit, but I can change them as you suggested if you insist.

netwerk/protocol/http/UserAgentOverrides.jsm
JustOff: Because it will be called later from `buildOverrides()`.

netwerk/protocol/http/UserAgentOverrides.jsm
JustOff: Each time the dynamic rules are updated, we need to rebuild the static rule map so that the default rules, which could be excluded from `gOverrides` during the previous update, are brought back into action.

netwerk/protocol/http/UserAgentOverrides.jsm
wolfbeast: Ah okay so IIUC gOverrides is already filled with the default set _and_ the dynamic overrides before getting to this code point?  In that case this would work since you'd want to avoid overriding dynamic with default. Could do with a short code comment explaining this though because it's absolutely _not_ obvious from the code itself.

Also, yes I do think `!gUpdatedOverrides` is what you want here, covering `null`, `undefined` and empty-string in one go, all of which you'd want in this condition.

JustOff: > gOverrides is already filled with the default set and the dynamic overrides before getting to this code point?

No, `gOverrides` never contains dynamic rules, it's only for static ones (both user-set and pre-set), and at this point we build `gOverrides` from scratch, based on data from `about:config` and taking into account the contents of `gUpdatedOverrides`.

> Could do with a short code comment explaining this though because it's absolutely not obvious from the code itself.

Ok, I will try to clarify this in a comment.

> Also, yes I do think `!gUpdatedOverrides` and `!gUpdatedOverrides.get(domain)` is what you want here, covering `null`, `undefined` and empty-string in one go

An empty string cannot be there, but ok, I will simplify the syntax as you ask.

> You should probably make it an and condition, grouped before negating, to clarify what you're doing here.

Sorry, but it just doesn't fit in my head. Let's better leave the "or" condition, and the logic will be explained in the comment above.

mattatobin: Prefs can have empty strings?

JustOff: I agree, but `gUpdatedOverrides` can only be an object, undefined or null and I can't imagine why anyone might need to dynamically send an empty string as an SSUAO rule for some domain. But anyway, I already agreed with a simplified syntax which covers that.

JustOff: I force-pushed all the changes we discussed here.

netwerk/protocol/http/UserAgentOverrides.jsm
```diff
@@ -159,15 +164,24 @@ function buildOverrides() {
   if (!Services.prefs.getBoolPref(PREF_OVERRIDES_ENABLED))
     return;
 
-  let builtUAs = new Map;
```
JustOff: This dead variable is not used anywhere.

netwerk/protocol/http/UserAgentOverrides.jsm
wolfbeast: Sorry but if `gOverrides` doesn't contain anything before this routine is called then I don't think this works because if there is no dynamic override and no user pref set (meaning just the default suppled one exists), then L176-180 will never be called and the override will never be set. That will effectively kill static SSUAOs

wolfbeast: hang on, no, sorry You are right. I'm just not getting this logic right in my head because of the negations and ORs.

JustOff: If there are no dynamic overrides then the condition is already satisfied (`!gUpdatedOverrides` is true) and L176-180 are called immediately.

JustOff: > this will build the static list containing defaults, unless there are dynamic updates for the domain in which case it's skipped, unless the pref is user-set. 

Absolutely. Please check the description I added [here](https://github.com/MoonchildProductions/UXP/commit/744c00d266bcc50cc875d9d389778a8cd937b093#diff-69a1f8fa8cb7bb1cf8179101d8f7c008R169-R184) in the force-pushed branch.

JustOff: > It gets confusing since the one if statement handles both exceptions, and exceptions of exceptions.

I read it like this: let's cycle trough the all prefs from `about:config` and select ones that satisfy the condintion (`not overriden dynamicaly` || `user-set`).

wolfbeast: I'd suggest, as I stated before, to clarify this logic by actually grouping it the way you read it:

``` javascript
!(gUpdatedOverrides && gUpdatedOverrides.get(domain)) ||
gPrefBranch.prefHasUserValue(domain)
```

That is a clear and direct translation of your 2 paragraphs "I read it like this"

JustOff: Done.

PR #1487, opened 2020-03-17
Make UXP applications capable of using SQLite for NSS instead of DBM

wolfbeast:
This adds a build-time configuration option to use the SQLite back-end for NSS certificate and key stores, adjusting PBKDF rounds for master passwords accordingly and making initialization in SQL mode such that it avoids an `slq_config` race.

This should automatically convert DBM to SQLite (if DBM support is built) upon first run, generating `key4.db` and `cert9.db` from the previous versions if present. That migration path is inside NSS.

Conversion should be tested. During development, operation in the new configuration has been tested with and without master password.
To enable, add the ac option `--enable-nss-sqlstore` to mozconfig. Incompatible combinations (`--disable-dbm` combined with `--disable-nss-sqlstore`, resulting in no db support) should error during configure.


wolfbeast:
Addressed the suggestion in part 4

Issue #1473, opened 2020-03-05
Improve SSUAO to provide more accurate OS information


JustOff:
As discussed [here](https://github.com/MoonchildProductions/Pale-Moon/issues/1733#issuecomment-595282189), we can rewrite ~~the static~~ UA overrides ~~using regexes~~ to more accurately reflect the OS.

wolfbeast:
Moving this to UXP since its platform work and impacts everyone.

JustOff:
So far, I have finished implementing support for the `%OS_SLICE%` macro, as we discussed above, but along the way I also found two more issues regarding SSUAO and dynamic rules update.

The first and most problematic is that all static SSUAO rules always take precedence over those received dynamically, which does not allow using dynamic updating to adjust builtin static rules, but only to add new ones. This can be fixed if we give dynamic rules priority over static rules from the default preference branch, but keep the highest priority for static rules created by the user.

The second is that dynamic rules are always updated at the interval specified in the settings, regardless of restarting the browser. However, given the changes related to the `%OS_SLICE%` macro, it would be better to schedule dynamic updates after a short delay each time the browser starts.

I have already prepared solutions to both of these additional issues and would like to ask how to organize this better. Do I need to create separate issues or can I refer to this one? Do I need to submit three separate pull requests or can I use a single one containing three commits?

mattatobin:
I want to know exactly what you are doing.

wolfbeast:
Please keep this issue focused only on implementing support for the %OS_SLICE% macro!

The other two issues can be looked at together in a single new issue for improving dynamic UA update handling, but don't belong here.
There's a reason I hid several comments because it was already branching off in several things and not remaining focused.

JustOff:
> I want to know exactly what you are doing.

Just as agreed [above](https://github.com/MoonchildProductions/UXP/issues/1473#issuecomment-595348580), details will follow.

> Please keep this issue focused only on implementing support for the %OS_SLICE% macro!

Ok.

mattatobin:
I would like details now please on how you intend to implement this. You can fully explain it or push up a branch somewhere and show me.

JustOff:
Well, [here](https://github.com/JustOff/UXP/commits/_ssuao_os_slice) it is. I am ready to answer your questions.

mattatobin:
Why did you refactor this? Why did you need to change anymore than JUST adding the string replace?

It looks like you are doing more than the scope of what is required.

[...]

PR #1494, opened 2020-03-24
Fix building `%OS_SLICE%` on non-Windows platforms

JustOff:
Build `%OS_SLICE%` according the [logic](https://github.com/MoonchildProductions/UXP/blob/13a8e788810c006e964cc3a87c0ddb9ca1eb9114/netwerk/protocol/http/nsHttpHandler.cpp#L711-L731) from `nsHttpHandler.cpp`. 

Note, that `mCompatDevice` and `mDeviceModelId` are always empty in UXP (we should probably consider removing them from `nsHttpHandler.cpp` as well).

This need to be verified on Linux and OS X.

Tag #1473.


wolfbeast:
Merging this untested. will build unstable to have people either complain or not.

Issue #1497, opened 2020-03-26
Reimplement checking for un-preprocessed files in a cross-platform way


JustOff:
Check for unpreprocessed files on Windows was [broken](https://github.com/MoonchildProductions/UXP/commit/e51afbcc2fe7360bbcf5654f6e31752c48098ca0) when adding Solaris support in #1262.

As a result, unprocessed files on Windows are not reported, which can lead to errors.

JustOff:
@athenian200, brace expansion doesn't work on Windows, so you have to find another general solution or provide a specific one for Solaris only.

mattatobin:
How about remove that garbage and just not screw up?

JustOff:
How about not calling a useful check garbage?

mattatobin:
How about not assuming what you consider useful is actually useful. Just slows down the process.

JustOff:
Let me remind you that earlier we [agreed](https://github.com/MoonchildProductions/UXP/issues/428#issuecomment-393859246) that it was useful. As for the time, then, IMHO, an extra 10-15 seconds is not worth mentioning.

As for the fact that no one noticed, we are just lucky that this did not lead to trouble, as it was already more than once. And I absolutely don't understand why you suddenly decided to resist to bring this instrument back to life.

mattatobin:
It was a stressful time then and that issue doesn't have the full story of why I decided to stop obstructing it at that time.

tl;dr that was then this is now and your understanding is not required.

But sure, we can ask @wolfbeast to flip a coin.

JustOff:
I hope you have not forgotten how a small typo can break everything? So, I believe that it's simply unreasonable to refuse an almost free additional automatic check.

> your understanding is not required

Please don't start this again.

wolfbeast:
Right. Stop your bickering.

First off, what was put in is pretty much a dirty hack - while having some use, "just in case" it's also not essential if people actually build and test their changes because it would come out in the wash soon enough if preprocessor directives are in an unprocessed file since it _will_ break.

Secondly, JustOff, you indeed do not have the entire image of why this wasn't obstructed further, so please don't pretend you do. Tobin and I discuss a hell of a lot of things about the project outside of public space in IM that you aren't privy to.

Instead of unloading this on Athenian200, how about you come up with a solution for your original code that works in all operating systems we are currently targeting? If you can't or won't, then we can remove this since it's not working in its current incarnation.

JustOff:
[...]

PR #1504, opened 2020-04-01
Issue #65 - Fix unprocessed directives in WebRequest.jsm

JustOff:
Fix unprocessed directives in `WebRequest.jsm` left after https://github.com/MoonchildProductions/UXP/commit/6a3907dae474c7e1f31df13a26605278d3500a0d, see also https://github.com/MoonchildProductions/UXP/issues/1497#issuecomment-607389362.

### DIFF
--- a/toolkit/modules/moz.build
+++ b/toolkit/modules/moz.build
@@ -11,7 +11,6 @@ EXTRA_JS_MODULES += [
     'addons/WebNavigation.jsm',
     'addons/WebNavigationContent.js',
     'addons/WebNavigationFrames.jsm',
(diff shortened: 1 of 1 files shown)


wolfbeast:
If you think the check still needs to be fixed, then please fix it, in a solaris-compatible way.

JustOff:
I think this fix should also be uplifted to `28.9-platform`.

mattatobin:
Why was this not its own issue?

JustOff:
You mean this hotfix does not belong to #65? I thought it is so.

PR #1506, opened 2020-04-02
Re-implement an un-preprocessed file check

### LINKED ISSUE #1497: Reimplement checking for un-preprocessed files in a cross-platform way

mattatobin:
Resolves #1497

----
Ping: @JustOff @athenian200 @dbsoft


python/mozbuild/mozbuild/mach_commands.py
```diff
@@ -541,6 +541,74 @@ def build(self, what=None, disable_extra_make_dependencies=None, jobs=0,
...
+        ]
+        
+        # Preprocessor directives that we check for
+        pp_directives = [
+            'define',
+            'if',
+            'ifdef',
+            'ifndef',
+            'elif',
+            'elifdef',
```
JustOff, line 567: In the previous implementation, @wolfbeast [insisted](https://github.com/MoonchildProductions/UXP/pull/429#discussion_r192563551) that `if`, `ifdef`, `ifndef`, `elif` and `elifdef` should not be checked.

python/mozbuild/mozbuild/mach_commands.py
```diff
@@ -541,6 +541,74 @@ def build(self, what=None, disable_extra_make_dependencies=None, jobs=0,
...
+            'filter',
+            'include',
+            'literal',
+            'undef',
+            'unfilter'
+        ]
+
+        # We need to construct the find command to work properly on old msys used when building on Windows but still
+        # work acceptably on newer versions of find
+        pp_find = "{0} {1}/{2} -name '*.{3}'".format('find', self.topobjdir, 'dist/bin', "' -o -name '*.".join(pp_types))
```
JustOff, line 580: It is better to limit the scan to files only:

`pp_find = "{0} {1}/{2} -type f \( -name '*.{3}' \)".format('find', '/d/UXP-build-palemoon', 'dist/bin', "' -o -name '*.".join(pp_types))`

otherwise there are errors in sdterr, such as:

```
grep: /d/UXP-build-palemoon/dist/bin/browser/chrome/devtools/content/responsive.html: Invalid request code
grep: /d/UXP-build-palemoon/dist/bin/browser/chrome/devtools/modules/devtools/client/responsive.html: Invalid request code
```

python/mozbuild/mozbuild/mach_commands.py
```diff
@@ -541,6 +541,74 @@ def build(self, what=None, disable_extra_make_dependencies=None, jobs=0,
...
+        pp_find_css = "{0} {1}/{2} -name '*.{3}'".format('find', self.topobjdir, 'dist/bin', 'css')
```
JustOff, line 581: Ditto, just to be safe.

python/mozbuild/mozbuild/mach_commands.py
```diff
@@ -541,6 +541,74 @@ def build(self, what=None, disable_extra_make_dependencies=None, jobs=0,
...
+        pp_cmd_css = '{0} | {1}'.format(pp_find_css, pp_xargs_css)
+
+        # Run the final find command and look for stdin messages
+        pp_result = subprocess.Popen(pp_cmd, stdout=subprocess.PIPE, shell=True).communicate()
+        pp_result_css = subprocess.Popen(pp_cmd_css, stdout=subprocess.PIPE, shell=True).communicate()
+
+        # We should only get output if something matched
+        if pp_result[0] or pp_result_css[0]:
+            pp_result_list = pp_result[0] + pp_result_css[0]
+            pp_result_list = pp_result_list.replace(self.topobjdir + '/dist/bin/', '').split('\n')
```
JustOff, line 598: I think it's more convenient to see the full path to un-processed files, i.e. do not remove it here.

python/mozbuild/mozbuild/mach_commands.py
```diff
@@ -541,6 +541,74 @@ def build(self, what=None, disable_extra_make_dependencies=None, jobs=0,
...
+            pp_final_list = []
+
+            for result in pp_result_list:
+              pp_final_list += [ result.split(':')[0] ]
```
JustOff, line 602: I'm not sure about the directives, but it might be better to keep them too, just to immediately understand what to look for.

python/mozbuild/mozbuild/mach_commands.py
```diff
@@ -541,6 +541,74 @@ def build(self, what=None, disable_extra_make_dependencies=None, jobs=0,
...
+
+            pp_final_list = list(dict.fromkeys(pp_final_list))
+
+            print('\nWARNING: The following files in dist/bin may need to be preprocessed:\n')
```
JustOff, line 606: I can't find the link now, but earlier someone told me that there should be an error, not a warning, because un-processed directives really break the code.

python/mozbuild/mozbuild/mach_commands.py
mattatobin: Any false positives here can be resolved by making sure that files with comments using the preprocessor directive prefix to make sure space them.

`# If this is some thing that is a thing then do something` not `#if this is some thing that is a thing`

This is also why it is a warning and not an error, though it is a warning that if ignored BECOMES an error.

python/mozbuild/mozbuild/mach_commands.py
mattatobin: Right, I forgot about that.

python/mozbuild/mozbuild/mach_commands.py
mattatobin: .

python/mozbuild/mozbuild/mach_commands.py
mattatobin: Moonchild doesn't want the output if any to be overly verbose. The warning text states it is in dist/bin. Normally, if you are relying on this for work you are working on it will be obvious. Checking the dist/bin file its self isn't very useful so copypasta of the path isn't either.

python/mozbuild/mozbuild/mach_commands.py
mattatobin: As stated, Moonchild doesn't want it to be overly verbose.

python/mozbuild/mozbuild/mach_commands.py
mattatobin: See previous comments.

python/mozbuild/mozbuild/mach_commands.py
JustOff: I'm rather on your side here, I just had to remind you that @wolfbeast strictly insisted on a different approach.

python/mozbuild/mozbuild/mach_commands.py
JustOff: Not that this is so important, but the lack of a full path here also seems inconsistent with the `no preprocessor directives found` warnings where it is. Perhaps you could change your mind on this?

mattatobin: I would, if it was spitting out the source file. It isn't so not very useful as far as the path is concerned. As for what it matches, Moonchild said he didn't want it to be overly verbose.

> [20200402.0924-11] \<NewTobinParadigm> http://personal.mattatobin.com/image/capture/94242413-7e6d-4a35-ada7-c90d20f11ccf.jpg
> [20200402.0924-35] \<NewTobinParadigm> yes i know the false positive.. there are only 2 files in the entire allience that will trigger it.. and I am going to fix it
> [20200402.0925-02] \<NewTobinParadigm> but I want line numbers for matches which means checking for all of them

[...]

> [20200402.1032-19] \<MoonchildPM> I personally would prefer it to be quieter than that. don't need debugging of the script.
> [20200402.1032-58] \<NewTobinParadigm> line numbers are useful
> [20200402.1033-14] \<MoonchildPM> when it actually finds something that errors, sure
> [20200402.1033-34] \<MoonchildPM> but otherwise it needs to just do its job and not spew output when things are OK.
> [20200402.1033-39] \<NewTobinParadigm> well it is going to print out the directives anyway cause it is a grep
> [20200402.1034-00] \<NewTobinParadigm> it doesn't output anything if there are no matches
> [20200402.1059-38] \<NewTobinParadigm> MoonchildPM: there http://personal.mattatobin.com/image/capture/fdf4614e-d625-4509-821a-30f6632eb488.jpg
> [20200402.1059-42] \<NewTobinParadigm> is that better?
> [20200402.1107-59] \<MoonchildPM> LGTM

mattatobin:
@JustOff I have redone the entire thing again this time in python form.

athenian200:
This works very well on Solaris. This solution seems superior to what we had before, definitely better than what I came up with while working on Solaris.

build/ppCheck.py
```diff
@@ -0,0 +1,78 @@
...
+
+from __future__ import print_function
+import os, sys
+
+print("\nChecking for un-preprocessed files...", end = ' ')
+
+if not sys.argv[1]:
+  print("\nYou did not supply a path to search")
+else:
+  DIST_PATH = sys.argv[1]
```
Conan-Kudo, line 11: Wouldn't this be better served [by using `argparse`](https://docs.python.org/3/howto/argparse.html) and setting a positional argument for this? For example:

```python
import argparse

parser = argparse.ArgumentParser()
parser.add_argument('dist_path', help="Path to search")
args = parser.parse_args()

DIST_PATH = args.dist_path
```

mattatobin: There is only one argument. So no.

build/ppCheck.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+import os, sys
+
+print("\nChecking for un-preprocessed files...", end = ' ')
+
+if not sys.argv[1]:
```
JustOff, line 8: It's better to check for `len(sys.argv) < 2`, otherwise:

```
$ ./ppCheck.py

Checking for un-preprocessed files... Traceback (most recent call last):
  File "./ppCheck.py", line 8, in <module>
    if not sys.argv[1]:
IndexError: list index out of range
```

build/ppCheck.py
```diff
@@ -0,0 +1,78 @@
...
+  print("\nYou did not supply a path to search")
```
JustOff, line 9: And we need to exit here, or:

```
$ ./ppCheck.py

Checking for un-preprocessed files...
You did not supply a path to search
Traceback (most recent call last):
  File "./ppCheck.py", line 49, in <module>
    for root, directories, filenames in os.walk(DIST_PATH):
NameError: name 'DIST_PATH' is not defined
```

wolfbeast: Do we really need all these run-time checks for something that is internal to our use?

mattatobin: Not really, but I'll em anyway. Just a couple of lines.

mattatobin:
@JustOff Final check please

JustOff:
LGTM, and I like it is in pure Python, I obviously should have done so from the very beginning.

PR #1520, opened 2020-04-21
Fix build break if "--disable-wmf" option is used.

xieyubo:


mattatobin:
Who asked for this? Where is the Issue that goes with this? Why is the code being so badly mangled?

Sorry, but we have contribution guidelines for a reason and not only was this not requested, collides with other work, it is a complete drive-by mess that absolutly does takes the scope of cross-platform development or the complexity of dependent configurations into account.

Also, did I mention it was mangled?

wolfbeast:
Please see the links below in "Remember, contributions to this repository should follow its contributing guidelines and code of conduct."

Issue #80, opened 2018-03-18
Stop using unified compilation of sources.


wolfbeast:
Mozilla has, for a while, been using unified source compilation, which concatenates multiple source files together before feeding the result to the compiler.
The idea behind this is that is creates shorter build times and _theoretically_ allows the compiler to optimize more parts of the code.

There are, however, some **major** drawbacks to this that are considered detrimental to our work here:
- **Dependencies** can be missed; if the concatenation includes another source file linking in the dependency, (e.g. a header), this header inclusion can be missing without the compiler complaining about it. If code changes and as a result the concatenation changes, we can suddenly be presented with build failures without a clear cause in sources that previously compiled without error.
- It can create **code bias** for sources that happen to end up in the same block; this in turn can lead to timing differences, races, and performance gaps.
- **Debugging** is more **difficult**; the actual compiled source file is a transitional chunk, not the actual source file.
- The build is more **fragile**: it relies on extra steps performed, is sensitive to newlines at ends of files or not, etc.
- Some source files can't be built unified, others can, and whether one or the other is true can be difficult to discern.

There is also **no conclusive data** that the goal set, on our size of source tree, was actually achieved or not.

wolfbeast:
Some things to watch out for when working on this:
- If classes/structs seem to be missing, check for `using namespace` statements as well as includes.
- If running into undefined externals when linking, aside from getting the proper includes, be aware Mozilla employees have also sprinkled `constexpr` everywhere (because it's new and cool...). If this is used on e.g. a forward declared constructor in a `.h`, this can lead to undefined external symbols at link time in non-unified mode of the relevant implementation is listed as `constexpr` in the `.cpp` -- it works when the caller and declaration containing files are concat-ed, but not forward declared, because it will be evaluated and resolved at **compile-time** with `constexpr`. This little tidbit cost me 1.5 hours of scratching my head why the concated version worked but the separated version did not, while everything was included that was needed.

wolfbeast:
Proposed workflow for future deprot fixing to prevent build bustage on master:

![deprot](https://user-images.githubusercontent.com/3359132/57178661-57cfab00-6e74-11e9-956e-632b002b81cd.png)

wolfbeast:
@adeshkp @g4jc @trav90

mattatobin:
Sounds fantasic!

trav90:
Sounds like a great plan!

wolfbeast:
[...]

PR #1522, opened 2020-04-23
Deunify dom

wolfbeast:
This de-unifies the dom/ directory (with a few exceptions) and fixes a boatload of deprot.
Start of this was done on master directly but we agreed we shouldn't do this, so this is on a branch.

Needs checking in Linux and Mac for build bustage


wolfbeast:
Tag #80

PR #1526, opened 2020-04-30
De-unify layout

wolfbeast:
This is the high-priority de-unification of /layout part of #80 , blocking further WebComponents work.

/layout seems to be particularly fragile in terms of unified building, so it really needs to be built using other applications than Basilisk which I used for this work so far and on other operating systems to catch anything that Windows builds might not have brought to light.


mattatobin:
Ping @adeshkp @dbsoft @athenian2000

dbsoft:
I built with the current changes without an issue on Mac and Windows, both Pale Moon and Basilisk.  If that is what the ping was about.

PR #1532, opened 2020-05-05
Fix warning about inline nsINode::GetFlattenedTreeParentNodeForStyle being undefined

adeshkp:
This warning creates a lot of noise both on Mac and Linux.

Tag #457

### DIFF
--- a/dom/base/ElementInlines.h
+++ b/dom/base/ElementInlines.h
@@ -8,6 +8,7 @@
 #define mozilla_dom_ElementInlines_h
 
 #include "mozilla/dom/Element.h"
+#include "nsIContentInlines.h"
 #include "nsIDocument.h"
 
 namespace mozilla {


g4jc:
Thanks for looking into this, it's been a pet peeve of mine since it was introduced during `dom_work`.

Another way of suppressing it is: 
```
--- a/dom/base/nsINode.h
+++ b/dom/base/nsINode.h
@@ -958,7 +958,7 @@ public:
    * root element's primary frame, such as scrollbar elements created
    * by the root scroll frame.
    */
-  inline nsINode* GetFlattenedTreeParentNodeForStyle() const;
+  nsINode* GetFlattenedTreeParentNodeForStyle() const;
 
   /**
    * Get the parent nsINode for this node if it is an Element.

```

Not sure how important it is to declare `inline` for that or not, primarily used to optimize during compile from what I understand.

mattatobin:
It would help if someone posted the warning.

adeshkp:
Very frequent.
```
 1:53.75 Warning: -Wundefined-inline in /Users/super/Basilisk/platform/dom/base/nsINode.h: inline function 'nsINode::GetFlattenedTreeParentNodeForStyle' is not defined
 1:53.75 /Users/super/Basilisk/platform/dom/base/nsINode.h:961:19: warning: inline function 'nsINode::GetFlattenedTreeParentNodeForStyle' is not defined [-Wundefined-inline]
 1:53.75   inline nsINode* GetFlattenedTreeParentNodeForStyle() const;
 1:53.75                   ^
 1:53.75 /Users/super/builds/basilisk/dist/include/mozilla/dom/ElementInlines.h:31:25: note: used here
 1:53.75   nsINode* parentNode = GetFlattenedTreeParentNodeForStyle();
 1:53.75                         ^
 1:55.13 9 warnings generated.
```

mattatobin:
I'll do some test builds today.

mattatobin:
It looks like we already have it here https://github.com/MoonchildProductions/UXP/blob/master/old-configure.in#L1019 so wtf

Issue #1493, opened 2020-03-23
Clang: replacement function 'operator new' cannot be declared 'inline' 


dbsoft:
Building Pale-Moon, Basilisk, UXP generate numerous warnings on newer versions of clang.

According to this mozilla bug, this is done for performance reasons that we don't really want to remove, their solution was just to disable the warning when using recent versions of clang:

https://bugzilla.mozilla.org/show_bug.cgi?id=928808

However their patch uses configure macros that don't seem to exist in the Pale Moon/UXP source base. 

The clang in Xcode 7 does not generate these warnings, but  the clang in Xcode 10 and 11 do.

mattatobin:
Thing is, these ARE only warnings so not really something that must be dealt with any time soon. This would be part of #457 actually.

wolfbeast:
FTR the listed Mozilla bug was fixed in gecko 30, the follow-up (to also use the workaround in /js) was landed in gecko 40 -- this is old code we have had for a long time so I don't see why this would even come up. anymore. The info in the linked BZ bug at least is outdated and shouldn't apply to us anyway.

dbsoft:
The reason I linked the mozilla bug was to show we should add the -Wno-inline-new-delete when using recent versions of clang, it had some tests for it and adds it when necessary, and as I said it is no longer in the code base.  mattatabin is right it should be in that other issue, I didn't notice that one but it does belong in there.

mattatobin:
If this is desired behavior we override it on the moz.build level and then only for what we need to not globally in ac or moz configure. Please stay away from either configure systems.

dbsoft:
Well, it looks like it is already in there, just not functioning with clang or with macos...  looks like the test is failing in config.log:

INFO: checking whether the C++ compiler supports -Wno-inline-new-delete... 
DEBUG: Creating `/var/folders/44/x_38dgxx70s1gbg8ptfq3fhc0000gn/T/conftest.NwUyVq.cpp` with content:
DEBUG: | 
DEBUG: | int
DEBUG: | main(void)
DEBUG: | {
DEBUG: | 
DEBUG: |   ;
DEBUG: |   return 0;
DEBUG: | }
DEBUG: Executing: `/usr/bin/clang++ -std=gnu++11 -Werror -Winline-new-delete -c /var/folders/44/x_38dgxx70s1gbg8ptfq3fhc0000
DEBUG: The command returned non-zero exit status 1.
DEBUG: Its error output was:
[...]

PR #1533, opened 2020-05-05
Supress warning about inlining `new` and `delete` in mozalloc

adeshkp:
This fixes a lot of unnecessary warning spew on Mac.

Same warning option is defined much below in the file but that never gets executed on Mac. Putting it here fixes the endless warnings about `new` and `delete` inline.

Tag #457

### DIFF
--- a/old-configure.in
+++ b/old-configure.in
@@ -798,6 +798,8 @@ case "$target" in
     MKCSHLIB='$(CC) $(CFLAGS) $(DSO_PIC_CFLAGS) $(DSO_LDOPTS) -o $@'
     MOZ_OPTIMIZE_FLAGS="-O3"
     CXXFLAGS="$CXXFLAGS -stdlib=libc++"
+    # Suppress the clang warning for the inline 'new' and 'delete' in mozalloc
+    CXXFLAGS="$CXXFLAGS -Wno-inline-new-delete"
     DLL_SUFFIX=".dylib"
     DSO_LDOPTS=''
     STRIP_FLAGS="$STRIP_FLAGS -x -S"


adeshkp:
Obviously you are the right person where this has to go, but I couldn't figure out other places where this had an effect. There are already mention of this warning at a few places but none of those seem to work. Both Linux and Mac are affected.

Like most frequent, more than a thousand times.

```
 4:34.17 Warning: -Winline-new-delete in /Users/super/builds/basilisk/dist/include/mozilla/mozalloc.h: replacement function 'operator delete[]' cannot be declared 'inline'
 4:34.17 /Users/super/builds/basilisk/dist/include/mozilla/mozalloc.h:224:21: warning: replacement function 'operator delete[]' cannot be declared 'inline' [-Winline-new-delete]
 4:34.17 MOZALLOC_EXPORT_NEW MOZALLOC_INLINE
 4:34.17                     ^
 4:34.17 /Users/super/builds/basilisk/dist/include/mozilla/mozalloc.h:39:27: note: expanded from macro 'MOZALLOC_INLINE'
 4:34.17 #  define MOZALLOC_INLINE MOZ_ALWAYS_INLINE_EVEN_DEBUG
 4:34.17                           ^
 4:34.17 /Users/super/builds/basilisk/dist/include/mozilla/Attributes.h:27:75: note: expanded from macro 'MOZ_ALWAYS_INLINE_EVEN_DEBUG'
 4:34.17 #  define MOZ_ALWAYS_INLINE_EVEN_DEBUG     __attribute__((always_inline)) inline
 4:34.17
```

mattatobin:
I'll investigate the places where this should be added if not where you slotted it in.

mattatobin:
It looks like we already have it here https://github.com/MoonchildProductions/UXP/blob/master/old-configure.in#L1019 so wtf

mattatobin:
After speaking with @dbsoft it seems that with Xcode 10 and up you need to specify using `libstdc++` because by default it uses `libc++` which fails all our warning tests.

Please see the newly written [build instructions for macintosh](http://developer.palemoon.org/build/mac/) for how to do that and align yourself to that configuration for `Pale Moon`.

As far as this PR goes.. Gonna `r-` it. We can look into seeing what can be done with the warning tests in the various configure systems and `libc++` bs at a later date or if it becomes necessary.. whichever comes first.

I do appreciate it though.

adeshkp:
The thing is on newer OS/Xcode you cannot even build without specifying stdlib as the compilation fails early. https://forum.palemoon.org/viewtopic.php?p=184676#p184676
So in that case question of warning does even arise.
That's why, of course, I've set `export CXXFLAGS=-stdlib=libc++` in the mozconfig and it's been like that for more than a year.

I'd say it's a generic compiler warning rather than a library issue. **In my second post, I pointed out that both Linux and Mac, i.e. gcc and clang, are affected.**

mattatobin:
It needs research. I don't know, a lot of us don't know. I think this needs properly fixed cause it is in warnings.mozconfig but isn't working properly on xcode-10.

dbsoft:
@adeshkp https://github.com/MoonchildProductions/UXP/issues/1493 is where I did my research.  I came up a couple of possible solutions... putting -stdlib=libc++ in CXX instead of CXXFLAGS fixes the warning tests. The easiest non-mozconfig fix would be to do a test to see if -stdlib=libc++ for darwin and add that to CXX there.

dbsoft:
I tested on Linux and did not get that warning either.

PR #1535, opened 2020-05-06
Remove TelemertyVFS

g4jc:
This reverts m-c Bug 668378 and completely removes Telemetry SQLite IO. As a bonus this fixes a potential crash in newer SQLite versions without the need for updating this useless telemetry shim.


wolfbeast:
I know there's been some discussion about this on IRC and that seems all favourable, but is this ready to merge as-is or did it need some more work?

g4jc:
It's ready to go. I finished testing it mailnews in both applications over a virtual file system (NFS share) just to be extra sure we still have functional VFS once telemetry was removed.

PR #1537, opened 2020-05-08
Better way to create display items for column backgrounds

### LINKED ISSUE #1355: Improve performance of table drawing again

win7-7:
Resolves #1355

Based to https://bugzilla.mozilla.org/show_bug.cgi?id=1388161 dropping parameter, also has build bustage fix for de-unified layout/xul. This makes applying 1409114 easier without requiring aDirtyRect/more changes to get the column backgrounds improvements to work.

PR has commit that changes nsDisplayListCollection() : to explicit nsDisplayListCollection(nsDisplayListBuilder* aBuilder) and add (aBuilder); where needed.

This change to explicit would be necessary to avoid error during compiling 1409114.

Also based to 1441796 - Part 1 and 3 and 1512244 - Part 1.

https://bugzilla.mozilla.org/show_bug.cgi?id=1441796

https://bugzilla.mozilla.org/show_bug.cgi?id=1512244

1512244 - Part 1 is included because it makes it easier to apply 1409114.

1441796 is in same commit because it comes chronologically before 1512244 - Part 1. Part 2 of 1441796 does not apply to UXP.

This PR also includes printing preview fix, based to https://bugzilla.mozilla.org/show_bug.cgi?id=1394249 as follow-up to commit https://github.com/MoonchildProductions/UXP/commit/6d504711a01e9be34725f9add22cc7aa311cb14f

Same as before 1409114 part 3 cannot be applied because it depends on Stylo/ArcSlice: https://bugzilla.mozilla.org/show_bug.cgi?id=1550554

Finally, last commit is based to: https://bugzilla.mozilla.org/show_bug.cgi?id=1409114

It also has AutoBuildingDisplayList changes for 1409114 part 11.

This is about 36-38 tiles per second improvement to Maze test. Now this is much closer to performance before issue #146. Maze test is still about 22-24 tiles less than -GL build before #146 was. It might be improved by optimizing layout and display list more than currently but those should be separate issues later since main regression is resolved after this and improvements to layout or display list need research. I tested this with New Moon build with Whole Program Optimization/-GL ( I always use -GL when building New Moon).

### DIFF
--- a/layout/base/nsDisplayList.cpp
+++ b/layout/base/nsDisplayList.cpp
@@ -864,10 +864,9 @@ nsDisplayListBuilder::FindAnimatedGeometryRootFor(nsDisplayItem* aItem)
 
 
 void nsDisplayListBuilder::MarkOutOfFlowFrameForDisplay(nsIFrame* aDirtyFrame,
-                                                        nsIFrame* aFrame,
-                                                        const nsRect& aDirtyRect)
+                                                        nsIFrame* aFrame)
 {
-  nsRect dirtyRectRelativeToDirtyFrame = aDirtyRect;
+  nsRect dirtyRectRelativeToDirtyFrame = GetDirtyRect();
   if (nsLayoutUtils::IsFixedPosFrameInDisplayPort(aFrame) &&
       IsPaintingToWindow()) {
     NS_ASSERTION(aDirtyFrame == aFrame->GetParent(), "Dirty frame should be viewport frame");
@@ -882,7 +881,9 @@ void nsDisplayListBuilder::MarkOutOfFlowFrameForDisplay(nsIFrame* aDirtyFrame,
       dirtyRectRelativeToDirtyFrame.SizeTo(aDirtyFrame->GetSize());
     }
   }
-  nsRect dirty = dirtyRectRelativeToDirtyFrame - aFrame->GetOffsetTo(aDirtyFrame);
+
+  nsPoint offset = aFrame->GetOffsetTo(aDirtyFrame);
+  nsRect dirty = dirtyRectRelativeToDirtyFrame - offset;
   nsRect overflowRect = aFrame->GetVisualOverflowRect();
 
   if (aFrame->IsTransformed() &&
@@ -1094,8 +1095,7 @@ nsDisplayListBuilder::ResetMarkedFramesForDisplayList()
 
 void
 nsDisplayListBuilder::MarkFramesForDisplayList(nsIFrame* aDirtyFrame,
-                                               const nsFrameList& aFrames,
-                                               const nsRect& aDirtyRect) {
+                                               const nsFrameList& aFrames) {
   for (nsIFrame* e : aFrames) {
     // Skip the AccessibleCaret frame when building no caret.
     if (!IsBuildingCaret()) {
@@ -1107,9 +1107,8 @@ nsDisplayListBuilder::MarkFramesForDisplayList(nsIFrame* aDirtyFrame,
         }
       }
     }
-    
     mFramesMarkedForDisplay.AppendElement(e);
-    MarkOutOfFlowFrameForDisplay(aDirtyFrame, e, aDirtyRect);
+    MarkOutOfFlowFrameForDisplay(aDirtyFrame, e);
   }
 }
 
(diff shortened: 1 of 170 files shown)


wolfbeast:
Tested building and performance with this PR using Basilisk, no build issues and maze performance is significantly better.
Also verified the known regressing website https://www.koutipandoras.gr/ isn't broken with this patch, so all looks good in that respect.

layout/generic/nsFrame.cpp
```diff
@@ -2637,12 +2632,15 @@ nsIFrame::BuildDisplayListForChild(nsDisplayListBuilder*   aBuilder,
...
-
   nsIAtom* childType = child->GetType();
   nsDisplayListBuilder::OutOfFlowDisplayData* savedOutOfFlowData = nullptr;
   bool isPlaceholder = false;
+
+  // dirty rect in child-relative coordinates
+  NS_ASSERTION(aBuilder->GetCurrentFrame() == this, "Wrong coord space!");
+  nsPoint offset = child->GetOffsetTo(this);
+  nsRect dirty = aBuilder->GetDirtyRect() - offset;
+
```
win7-7, line 2712: This should also not be split similar to nsDisplayList.cpp and instead be nsRect dirty = aBuilder->GetDirtyRect() - child->GetOffsetTo(this); and be moved upwards before nsIAtom*, correct?

layout/base/nsDisplayList.cpp
```diff
@@ -882,7 +881,9 @@ void nsDisplayListBuilder::MarkOutOfFlowFrameForDisplay(nsIFrame* aDirtyFrame,
       dirtyRectRelativeToDirtyFrame.SizeTo(aDirtyFrame->GetSize());
     }
   }
-  nsRect dirty = dirtyRectRelativeToDirtyFrame - aFrame->GetOffsetTo(aDirtyFrame);
+
+  nsPoint offset = aFrame->GetOffsetTo(aDirtyFrame);
+  nsRect dirty = dirtyRectRelativeToDirtyFrame - offset;
```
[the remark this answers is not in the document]
win7-7: I can change this back to what it was before. I took change across from 1388161. Mozilla probably did split as preparation for retained display lists that are not needed to UXP.

layout/base/nsDisplayList.h
```diff
@@ -1175,6 +1184,7 @@ class nsDisplayListBuilder {
     nsRect        mCaretRect;
     uint32_t      mFirstFrameMarkedForDisplay;
     bool          mIsBackgroundOnly;
+    nsIFrame*     mPresShellIgnoreScrollFrame;
     // This is a per-document flag turning off event handling for all content
```
win7-7, line 1199: This should be changed: nsIFrame*     mPresShellIgnoreScrollFrame; should be moved one line down to align with Mozilla version of file as mPresShellIgnoreScrollFrame being last after mInsidePointerEventsNoneDoc.

layout/base/nsDisplayList.cpp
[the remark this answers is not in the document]
wolfbeast: Avoiding unnecessary variables is a good thing ;)

layout/generic/nsFrame.cpp
wolfbeast: Avoiding the variable, yes. moving it up, no -- the added assertion might not behave the same there due to the `GetType()` call.

layout/base/nsDisplayList.h
wolfbeast: Not strictly necessary but if you're touching this more anyway, might as well.

wolfbeast:
In case you aren't aware: you can update this PR with the necessary changes by pushing additional commit(s) to the column-backgrounds-pr-2 branch on win7-7/UXP -- it'll update here automatically.

win7-7:
Ready to merge?

Last commit fixes nsDisplayList.cpp and nsFrame.cpp and changes PresShellState member nsIFrame*     mPresShellIgnoreScrollFrame; to be last.

PR #1554, opened 2020-05-18
Deunify netwerk

wolfbeast:
This de-unifies `/network` and subdirs with the exception of `/netwerk/protocol/websocket` because that's an entangled IPC mess that I wasn't able to figure out within a reasonable amount of time.

Tag #80 

Only tested with Basilisk/Win64. No energy to do the Linux carousel today, maybe y'all can help with that ;)


wolfbeast:
Shot off a build of Pale Moon/Linux64 gcc7 on the final branch state, looks good; merging

Issue #1356, opened 2020-01-16
Forms: Click() isn't executing properly when elements are defined disabled


wolfbeast:
There are two controls for elements being disabled/enabled which can confuse event handlers. `@disabled` is what is usually used to set an element disabled or not from js, `-moz-user-input: disabled/enabled` resulting from raw element properties.

We currently look at the `-moz-user-input` attribute; this is normally not manipulated by websites and can remain in the wrong state when enabling controls in forms through js, resulting in not registering the event the first time since the state is still incorrect during the event.

This is an old bug going back to at least Gecko/20. Mozilla solved this by removing `-moz-user-input: enabled/disabled` making everything use a single control. We should follow that.
Potential impact on the web would only be where websites would use Mozilla-prefixed css properties (which would have broken when Mozilla landed this) so that should be of no consequence.

Relevant BZ bug: [1405087](https://bugzilla.mozilla.org/show_bug.cgi?id=1405087)
Forum thread (title misnomer): https://forum.palemoon.org/viewtopic.php?f=29&t=23587#p181959

wolfbeast:
So you've done some work, and opened+closed a PR for this... what's going on here?

mattatobin:
It is slightly incomplete and collides with the big work going on.

PR #1559, opened 2020-05-21
Restore -moz-input-disabled and allow events to target disabled form controls.

### LINKED ISSUE #1557: Fieldset disabled no longer works
Regression 28.8.4 -> 28.9.0, still broken in 28.9.3

Potentially a regression from the html5 parser work done? Marking as parser component for now.

http://jsbin.com/bibiqi/1/edit?html,css,output -- test input fields should all have a red background in the disabled fieldset.

athenian200:
Resolves #1557.

This PR reverts the fix for #1356 which caused #1557 and solves the original problem in a different way. The patch is based on https://bugzilla.mozilla.org/show_bug.cgi?id=329509, which seems like a more direct (and less destructive) approach to solving the issue.

Several HTML elements are now handled via WidgetEvents rather than EventMessages, and the Click method in nsGenericHTMLElement is explicitly changed to allow events to pass through. Previously, changing certain disabled form states back to enabled did nothing because disabled meant it also wouldn't process the event re-enabling the control.

I'm not sure whether it's okay to include reverting my original changes as part of this PR or not, but I don't mind separating them out if that's preferred.


wolfbeast:
> I'm not sure whether it's okay to include reverting my original changes as part of this PR or not, but I don't mind separating them out if that's preferred.

It is always OK to include a reversion of a previous commit in a PR that solves it another way. After all, the reversion is part of the fixing process. It just needs to have the original PR labelled with "backed out" once this is landed on master.

PR #1567, opened 2020-05-25
Reader Updates

Ascrod:
This is an update for the Reader component.

From here on I intend to update the Readability library files directly from the source instead of correlating with Firefox. This is because I noticed that some of Mozilla's automated codebase-wide style refactorings are touching the files specifically marked as "DO NOT MODIFY". Nothing breaking yet as far as I know, but it's not a risk I want to introduce.

### DIFF
--- a/toolkit/components/reader/AboutReader.jsm
+++ b/toolkit/components/reader/AboutReader.jsm
@@ -70,8 +70,6 @@ var AboutReader = function(win, articlePromise) {
 
   Services.obs.addObserver(this, "inner-window-destroyed", false);
 
-  doc.addEventListener("visibilitychange", this);
-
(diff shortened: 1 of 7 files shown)


mattatobin:
Ascrod please use either the UXP or normalized mozilla bug commit style.

See UXP contribution guidelines for UXP style and commit history for the normalized mozilla bug style. Then force push the revised commits to your pr branch.

Ascrod:
Aye aye, captain.

mattatobin:
Does this require any fe changes to `Basilisk`?

Ascrod:
Not really. The Android cleanup removes a little bit of unused UI, but nothing we care about and nothing that will affect themes. I checked.

Issue #1564, opened 2020-05-24
Create nsIdentifierMapEntry.h because ShadowRoot.h needs the class


g4jc:
As an [m-c dev put it](https://hg.mozilla.org/mozilla-central/rev/621ec6b67f20):
> `nsIdentifierMapEntry.h` is separated from `nsDocument.h` because `ShadowRoot.h` needs the class but exposing `nsDocument.h` to the global and includes it causes bustage on Linux.

As for m-c, they did a number of intensive refactors at this point. Including, but not limited to:
* An optimization rewrite of nsIdentifierMap. That supposedly increased speed, yet caused further regressions. (Bug 1355787, Bug 1577560, Bug 1217436)
* Removed all use of nsI* DOM features, since they are were no longer needed (addons used them). (Bug 1332812)

As such, we can't do any direct porting here. But we can follow a similar path to break up the header file. The above also blocks [Bug 1426494](https://bugzilla.mozilla.org/show_bug.cgi?id=1426494) - Share more code between nsIDocument and ShadowRoot. (That will also need further work in a separate bug).

A [simple 1:1 cut and paste patch available](https://paste.debian.net/plain/1148625), but it does not work.
Splitting the header results in the following error:
```
 0:38.90 gmake[2]: Circular /UXP/mozilla/CLOBBER <- /UXP/mozilla/CLOBBER dependency dropped.
 0:51.85 liburiloader_prefetch.a.desc
 0:55.78 libtoolkit_components_places.a.desc
 1:44.26 liblayout_mathml.a.desc
 1:53.08 libdom_xslt_xpath.a.desc
 1:59.05 In file included from /UXP/obj-x86_64-pc-linux-gnu/dist/include/nsISupportsUtils.h:14:0,
 1:59.05                  from /UXP/obj-x86_64-pc-linux-gnu/dist/include/nsISupports.h:130,
 1:59.05                  from /UXP/obj-x86_64-pc-linux-gnu/dist/include/nsCycleCollectionTraversalCallback.h:12,
 1:59.05                  from /UXP/obj-x86_64-pc-linux-gnu/dist/include/nsCycleCollectionNoteChild.h:13,
 1:59.05                  from /UXP/obj-x86_64-pc-linux-gnu/dist/include/nsCycleCollectionParticipant.h:12,
 1:59.05                  from /UXP/obj-x86_64-pc-linux-gnu/dist/include/mozilla/EditTransactionBase.h:9,
[...]

PR #1568, opened 2020-05-25
Split off nsIdentifierMapEntry in its own header

wolfbeast:
+ Fix dependency fallout from removing nsDocument.h from ShadowRoot.h
Issue #1564 

@g4jc this should do a bit better for building, I think.


wolfbeast:
Update incoming for debug builds.

wolfbeast:
Assuming that is the include you meant. You were kind of ambiguous

g4jc:
My apologies as I had not seen your previous debug fixes prior to my review comment. Hence last commit is not needed. The first two commits  you did work great, tested several times on different settings.

PR #1576, opened 2020-06-02
Fix the wrong position when we calculate the position for position:absolute child

### LINKED ISSUE #1575: Position absolute does not work as expected in table with border collapse
https://bugzilla.mozilla.org/show_bug.cgi?id=1379306

As reported on forum: https://forum.palemoon.org/viewtopic.php?f=29&t=24504

Steps to reproduce:

https://codepen.io/vr8088/pen/yLeBVbN

The table is set to "border-collapse: collapse"
A child td has "position: relative" and a non-zero border width
The td contains an element set to "position: absolute"

Expected result: The element with "position: absolute" should be positioned relative to the position of the td (= of the first not-static ancestor element).

Actual result: It is positioned relative to the td plus a small offset in the x and y directions. The size of this offset seems to depend on the td's border width.

Also testcases from 1379306, all of those will work after applying patch from 1379306.

I will do PR to resolve this.

win7-7:
Resolves #1575 

Based to https://bugzilla.mozilla.org/show_bug.cgi?id=1379306

Steps to reproduce:

https://codepen.io/vr8088/pen/yLeBVbN

The table is set to "border-collapse: collapse"
A child td has "position: relative" and a non-zero border width
The td contains an element set to "position: absolute"

Results before PR: The element is positioned relative to the td plus a small offset in the x and y directions. The size of this offset seems to depend on the td's border width.

Results with PR: The element with "position: absolute" is positioned relative to the position of the td (= of the first not-static ancestor element).

Also testcases from 1379306, all of those do work with this PR.

@wolfbeast can review this.

### DIFF
--- a/layout/generic/nsAbsoluteContainingBlock.cpp
+++ b/layout/generic/nsAbsoluteContainingBlock.cpp
@@ -658,8 +658,8 @@ nsAbsoluteContainingBlock::ReflowAbsoluteFrame(nsIFrame*                aDelegat
 
   // Get the border values
   WritingMode outerWM = aReflowInput.GetWritingMode();
-  const LogicalMargin border(outerWM,
-                             aReflowInput.mStyleBorder->GetComputedBorder());
+  const LogicalMargin border(outerWM, aDelegatingFrame->GetUsedBorder());
+
   LogicalMargin margin =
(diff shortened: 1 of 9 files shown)


wolfbeast:
Please keep PR comments specific to what the PR does in the future. Testcases, STR, etc. belong solely in the issue description. I've updated your PR comment accordingly.

PR #1588, opened 2020-06-11
Remove no longer existing test headers from WebIDL

g4jc:
Although the headers were removed, they are still referenced in WebIDL causing compiler to fail for debug builds.


mattatobin:
Fix the format of the `moz.build` file. As-is it is unacceptable.

g4jc:
This patch simply removes `'TestFunctions.webidl'` and `TestInterfaceJSMaplikeSetlikeIterable.webidl`.

Format was unchanged - Unless you mean the white space after `WEBIDL_FILES += [`?

mattatobin:
Do as you are instructed.

g4jc:
Whitespace nit fixed and documented in commit.

Issue #1312, opened 2019-12-03
MailNews Core Enhancements


mattatobin:
**MailNews:**
- [ ] [1522608](https://bugzilla.mozilla.org/show_bug.cgi?id=1522608) - remove uses of eval() (Can take some of this patch, not all)
- [ ] [1461106](https://bugzilla.mozilla.org/show_bug.cgi?id=1461106) - Remove SMTP password from cache when deleted from password manager to prevent stale connection attempts.
- [ ] [1519387](https://bugzilla.mozilla.org/show_bug.cgi?id=1519387) - Implement IMAP cache logging.
- [ ] [1195026](https://bugzilla.mozilla.org/show_bug.cgi?id=1195026) - improve error message on SMTP errors 5.7.0 and 5.7.1.
- [ ] [144876](https://bugzilla.mozilla.org/show_bug.cgi?id=1448476) - Port bug 1448476 (Cache entry corruption after writing the alternate data.) to mailnews (IMAP, NNTP). (_UXP Change Required If We Take This_)

**Oauth2**
- [ ] [1597933](https://bugzilla.mozilla.org/show_bug.cgi?id=1597933) - clean up OAuth2 code
- [ ] [1599054](https://bugzilla.mozilla.org/show_bug.cgi?id=1599054) - allow callers to ommit sending OAuth2 client_secret parameter. 
- [ ] [1453643](https://bugzilla.mozilla.org/show_bug.cgi?id=1453643) - Enable proper retry on oauth2 authenication failure.

**NSS:**
- [ ] [1529222](https://bugzilla.mozilla.org/show_bug.cgi?id=1529222) - Disable support for MD2 in S/MIME multipart; nsCMSMessage should use HASH_GetHashOidTagByHashType. 
- [ ] [1529228](https://bugzilla.mozilla.org/show_bug.cgi?id=1529228) - Export NSS CMS functions required by TB/SM in mailnews/nss-extra.symbols.
 
**SMIME**
- [ ] [1528615](https://bugzilla.mozilla.org/show_bug.cgi?id=1528615) - regression fix for opaque S/MIME signatures. (See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1011625#c15)
- [ ] [1509586](https://bugzilla.mozilla.org/show_bug.cgi?id=1509586) - Look for requireEncryptMessage in params.smFields in msgCompSecurityInfo.js.

g4jc:
[...]

PR #1590, opened 2020-06-11
Don't allow STARTTLS when server sends PREAUTH 

g4jc:
Tag #1312

### DIFF
--- a/mailnews/imap/src/nsImapProtocol.cpp
+++ b/mailnews/imap/src/nsImapProtocol.cpp
@@ -1526,28 +1526,44 @@ void nsImapProtocol::EstablishServerConnection()
   }
   else if (!PL_strncasecmp(serverResponse, ESC_PREAUTH, ESC_PREAUTH_LEN))
   {
-    // we've been pre-authenticated.
-    // we can skip the whole password step, right into the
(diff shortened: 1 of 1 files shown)


mattatobin:
And this is? What is it for. Why is it here.

wolfbeast:
I'm still in the process of getting sec bug access.

wolfbeast:
The following was disclosed in public e-mail to me:

>"There isn't much in the bug, besides discussion how to fix it best. Basically a MITM could send PREAUTH to the client, causing the client to skip STARTTLS, and then the server would get to see the emails that the client is willing to save on the server, such as draft messages."

So basically STARTLS was previously ignored, but if used now aborts if PREAUTH is sent to err on the side of caution, IIUC.

No further action has been taken by the thunderbird team so far re: bug access. We can probably land this as-is without verification.

mattatobin:
Thank you @wolfbest

wolfbeast:
I've been given bug access, read through the discussion there and I can affirm that the solution landed is correct and covers all potential abuse scenarios (including passive leaking of data over unencrypted connections without MitM)

mattatobin:
Thank you @wolfbeast

PR #1591, opened 2020-06-13
Implement ShadowRoot.activeElement

mattatobin:
It builds and `<style scoped>` works but the devirtualizing that was done is still in question. I am not qualified to evaluate the repercussions.

Therefor, I request @wolfbeast to make that determination.


mattatobin:
Relevant Info:

> Google WebComponents is at its base completely incompatible with Gecko-based things. GWC uses strings where we use atoms. So it needs a flexible class that can handle both, and a flexible structure that can handle mozilla-style document roots and google-style shadowdom tree roots.
> 
> So JS needs a terribly ambiguous interface to be able to talk to what we have and what GWC needs us to add or in layman's terms: square peg, round hole.

Issue #1578, opened 2020-06-06
Add global menubar support for GTK


Lootyhoof:
For several years, a few desktop environments on *nix have supported a global menubar similar to OS X (initially Ubuntu's Unity, but later also XFCE/MATE via [vala-panel-appmenu](https://github.com/rilian-la-te/vala-panel-appmenu) and KDE via its included Global Menu plasmoid). Many applications, both GTK and Qt, are able to export their menubars such that they can support this. Historically however, Gecko-based applications haven't been able to do this. However, the Ubuntu-specific maintainers for Firefox have continually maintained a patchset to enable this, and included it in their stable builds of Firefox. Recently, I decided to take a look at the [patch](https://bazaar.launchpad.net/~mozillateam/firefox/firefox.xenial/view/1187/debian/patches/unity-menubar.patch) aimed at a Gecko version close to our fork point, and after applying it and adjusting for the current state of our tree, it works as expected:

![Pale Moon, with global menu patch](https://i.imgur.com/vtNcbQt.png)

This does require quite a bit of addition to the GTK widget code to enable this. I've actually already done this, and cleaned it up to suit the UXP coding style, which can be found over [here](https://github.com/Lootyhoof/UXP/commit/1d98ca2fe5b09f0b8f07053c72e5714b92b64755). I can confirm it compiles as expected.

Just to note also that this does NOT change any default behaviour; it's pref-controlled, and disabled by default. It does require a few additional packages installed for those that want it, but it should be noted that if those packages are not installed (and/or the menubar applet is not enabled in the DE) and the pref is enabled, nothing happens and the menubar stays in the window as before.

Let me know what you think, and if you'd like me to submit a PR (I can adjust the commit message as required too). It does also require some minor patches application-side too, as can be seen in the original Ubuntu patch, though I'm happy to submit issues/PRs as required there too if we want to pursue this.

mattatobin:
I think this would need to be conditional at build time unless you are going to modify all applications.

Also, what about GTK2 vs 3 and for applications like mail which have multiple windows with menu bars not just the main window?

Lootyhoof:
>unless you are going to modify all applications.

Yes, the idea is to make this available to all consumers of UXP, disabled behind the pref `ui.use_global_menubar`. Users that want it could enable it on an application of their choosing. I've already mentioned that I'm willing to submit the patches.

>what about GTK2 vs 3

Both are supported by this patch, and I've personally compiled on both configurations to confirm. 

[...]

PR #1595, opened 2020-06-14
Remove unnecessary MOZ_ASSERT from widget/gtk/nsNativeMenuService.cpp

Lootyhoof:
Tag #1578 

The `MOZ_ASSERT` was tripping in debug builds, crashing the browser. This isn't included in older versions of the Ubuntu patchset, nor does it cause any issues by NOT having it, so to avoid stability issues it has been removed.


mattatobin:
If it crashes we will simply have to fix the crash, etc.

PR #1613, opened 2020-07-28
[Image/CSS] Intrinsic Aspect Ratio

### LINKED ISSUE #1620: Use Intrinsic Aspect Ratio for Images
Intrinsic Aspect Ratio support:

> If img's width and height attribute values, when parsed using the rules for parsing dimension values, are both not an error, not a percentage, and non-zero, then use the ratio resulting from dividing the width attribute value by the height attribute value.

These changes allow websites to specify the dimensions of an image via the HTML `width` and `height` attributes as well as CSS-based sizes, resulting in a ratio-preserved scale frame for images before they've loaded, preventing website reflow on dynamically sized images (most often by setting one dimension to `auto`). This only applies to image frames at present. If the image is loaded, the ratio defers to the actual image dimensions. If the image is unable to load, cross-browser-compatibility is retained by collapsing the frame.

This code includes changes to pave the way for css aspect-ratio. If it goes too far, please let me know. Particularly, `layout/style/nsCSSPropList.h` and `nsRuleNode.cpp` are not part of the Firefox code at all. I don't know if they're necessary for the existing limited implementation, but it seemed like the safest path to add AspectRatio to them.
Additionally, if any or all of the tests are not required, they can be omitted from the final commit.

Ref: BZ [Bug 1547231](https://bugzilla.mozilla.org/show_bug.cgi?id=1547231), BZ [Bug 1559094](https://bugzilla.mozilla.org/show_bug.cgi?id=1559094), BZ [Bug 1633434](https://bugzilla.mozilla.org/show_bug.cgi?id=1633434), BZ [Bug 1565690](https://bugzilla.mozilla.org/show_bug.cgi?id=1565690), and BZ [Bug 1602047](https://bugzilla.mozilla.org/show_bug.cgi?id=1602047)
Also possibly eventually, BZ [Bug 1621333](https://bugzilla.mozilla.org/show_bug.cgi?id=1621333)

Resolved in [PR 1613](https://github.com/MoonchildProductions/UXP/pull/1613)

RealityRipple:
Intrinsic Aspect Ratio support:

> If img's width and height attribute values, when parsed using the rules for parsing dimension values, are both not an error, not a percentage, and non-zero, then use the ratio resulting from dividing the width attribute value by the height attribute value.

These changes allow websites to specify the dimensions of an image via the HTML `width` and `height` attributes as well as CSS-based sizes, resulting in a ratio-preserved scale frame for images before they've loaded, preventing website reflow on dynamically sized images (most often by setting one dimension to `auto`). This only applies to image frames at present. If the image is loaded, the ratio defers to the actual image dimensions. If the image is unable to load, cross-browser-compatibility is retained by collapsing the frame. Resolves #1620.

### Actions taken:
Calculate missing width or height based on Aspect Ratio.
Use HTML-set width and height to determine Aspect Ratio before image is loaded.

### Preferences Added:
layout.css.intrinsic-aspect-ratio.enabled [default: true]

### Original Mozilla Changes:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1547231
- https://bugzilla.mozilla.org/show_bug.cgi?id=1559094
- https://bugzilla.mozilla.org/show_bug.cgi?id=1633434
- https://bugzilla.mozilla.org/show_bug.cgi?id=1565690
- https://bugzilla.mozilla.org/show_bug.cgi?id=1602047

### Additional Notes:
`layout/style/cssPropList.h` and `layout/style/nsRuleNode.cpp` are complete guesses on my part.

### DIFF
--- a/dom/html/HTMLImageElement.cpp
+++ b/dom/html/HTMLImageElement.cpp
@@ -325,7 +325,7 @@ HTMLImageElement::MapAttributesIntoRule(const nsMappedAttributes* aAttributes,
   nsGenericHTMLElement::MapImageAlignAttributeInto(aAttributes, aData);
   nsGenericHTMLElement::MapImageBorderAttributeInto(aAttributes, aData);
   nsGenericHTMLElement::MapImageMarginAttributeInto(aAttributes, aData);
-  nsGenericHTMLElement::MapImageSizeAttributesInto(aAttributes, aData);
+  nsGenericHTMLElement::MapImageSizeAttributesInto(aAttributes, aData, true);
   nsGenericHTMLElement::MapCommonAttributesInto(aAttributes, aData);
 }
 
--- a/dom/html/HTMLImageElement.h
+++ b/dom/html/HTMLImageElement.h
@@ -196,6 +196,11 @@ public:
     return GetReferrerPolicyAsEnum();
   }
 
+  bool IsAwaitingLoad() const
+  {
+    return !!mPendingImageLoadTask;
+  }
+
   int32_t X();
   int32_t Y();
   // Uses XPCOM GetLowsrc.
--- a/dom/html/nsGenericHTMLElement.cpp
+++ b/dom/html/nsGenericHTMLElement.cpp
@@ -1480,7 +1480,7 @@ nsGenericHTMLElement::MapImageSizeAttributesInto(const nsMappedAttributes* aAttr
     }
   }
 
-  if (Preferences::GetBool("layout.css.width-and-height-map-to-aspect-ratio.enabled") &&
+  if (Preferences::GetBool("layout.css.intrinsic-aspect-ratio.enabled") &&
       aMapAspectRatio && aWidth && aHeight) {
     Maybe<double> w;
     if (aWidth->Type() == nsAttrValue::eInteger) {
--- a/dom/html/nsGenericHTMLElement.cpp
+++ b/dom/html/nsGenericHTMLElement.cpp
@@ -1449,29 +1449,59 @@ nsGenericHTMLElement::MapImageMarginAttributeInto(const nsMappedAttributes* aAtt
 
 void
 nsGenericHTMLElement::MapImageSizeAttributesInto(const nsMappedAttributes* aAttributes,
-                                                 nsRuleData* aData)
+                                                 nsRuleData* aData,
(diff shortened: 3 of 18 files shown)


wolfbeast:
Also... this needs an Issue with these details in it. This PR as-is won't be accepted without an attached issue because it is not minor and changes behavior.

RealityRipple:
> Either list the bugs or link to mercurial.

Done

> Also... this needs an Issue with these details in it. This PR as-is won't be accepted without an attached issue because it is not minor and changes behavior.

Since this is a draft, I was planning to write a new PR using the master branch after the code is ready.

wolfbeast:
> Since this is a draft, I was planning to write a new PR using the master branch after the code is ready.

Of course, but please file an issue in the issue tracker with the details, link your commits to it with the established commit message format we use, so this can be found later with relevant searches.

RealityRipple:
Today's Task: `IsAwaitingLoad()`

RealityRipple:
Been testing the changes using a merged release branch, everything's handling okay so far. I'd still kind of prefer the preference to be false for at least one or two versions, regardless, of course, but I think everything's working. Whether everything's _right_ or not is a different matter, of course.

If everything does end up looking right, let me know before you merge. I'd like to get rid of my personal comments in this and my previous pull request before then. No need to clutter up the source.

wolfbeast:
I'm building a try version with the changes in this PR to give it a test today.

wolfbeast:
I've applied this to master, and ran your test, but it doesn't seem correct. Looks like it's doing exactly the same as -release does and there is no change.
Image1

RealityRipple:
Did you turn the preference on in about:config?

wolfbeast:
ohh you kept the pref off. sorry I missed that.

wolfbeast:
Ugh, what a terribly long pref name! Can you maybe shorten it to something a bit more sane? We're not mozillaWithMaybeThisOrThatLongFunctionNames() here.

Flipping it indeed gives the expected results. As far as I'm concerned it's good to land and pref **on** by default; you won't get any usage of it otherwise for something that's relatively obscure. having a pref to be able to flip it off in case there are breaking issues is a good thing (for now anyway).

RealityRipple:
Right-o!

wolfbeast:
Thanks!

Issue #1619, opened 2020-07-31
Convert Intrinsic Ratio from nsSize to Float


RealityRipple:
Aspect Ratio handling simplified by using floating point integers:
- Multiplication of value (or inverse value) to a known side for Scaling
- No unequal equal values such as "4/3" vs "8/6" vs "20/15"
- Truly "Empty" aspect ratios, even if one dimension is not 0

This is accomplished by the creation of a new AspectRatio structure to replace nsSize usage, which is just some useful functions wrapped around a float.

Ref: BZ [Bug 1547792](https://bugzilla.mozilla.org/show_bug.cgi?id=1547792) and BZ [Bug 1548768](https://bugzilla.mozilla.org/show_bug.cgi?id=1548768)

Blocks [Issue 1620](https://github.com/MoonchildProductions/UXP/issues/1620)

Resolved in [PR 1622](https://github.com/MoonchildProductions/UXP/pull/1622)

RealityRipple:
I'm not certain, but BZ Bug 1548768 might need direct `aspect-ratio` CSS support before a testcase can be produced. It needs both sizes set as `auto` and a valid intrinsic ratio at the time `ComputeSizeWithIntrinsicDimensions` is called for the given nsFrame. I'm pretty sure the actual problem is fixed as `logicalRatio` is what values are `ApplyTo()`'d, which is what the Phabricator comments suggest doing, but without a testcase, I can't be sure. If it does all look good, we can just delete those comments and apply the patch for now, and come back to the testcase if `aspect-ratio` makes its debut.

wolfbeast:
...or write a testcase that specifically tests what you're supposed to be fixing here.

RealityRipple:
That's kind of part of the problem. I'm not sure exactly when `ComputeSizeWithIntrinsicDimensions` is called, and if the code in question ever gets run. Despite being about this change, the testcase itself would be reliant on the existence of `aspect-ratio` as a CSS property (which we're not doing), because even after #1620 is done, the internal aspect-ratio value will still need both a width and height set for a ratio to exist, which means the changed code won't run because both a width and height are set to something other than `auto`. I can write a testcase to make sure images and objects are the right size in both horizontal and vertical writing modes, both in general and upon their immediate creation, but I can't guarantee that's the right thing to test.

mattatobin:
I know this won't reflect your changes but you do know we have a Code Cross-Reference, right?

----
[...]

PR #1622, opened 2020-07-31
Convert Intrinsic Ratio from nsSize to Float

RealityRipple:
First of two-phase implementation of Intrinsic Aspect Ratios from `width=` and `height=` HTML values:

Any use of `nsSize` for Aspect Ratio values now uses a `float` value. Some special helper functions simplify basic aspect ratio tasks, both in current use and in phase two of Intrinsic Aspect Ratio support.

This code should cause no real functional changes as of yet. Resolves issue #1619.

### DIFF
--- a/layout/generic/nsFlexContainerFrame.cpp
+++ b/layout/generic/nsFlexContainerFrame.cpp
@@ -1529,9 +1529,9 @@ MainSizeFromAspectRatio(nscoord aCrossSize,
 {
   MOZ_ASSERT(aIntrinsicRatio,
              "Invalid ratio; will divide by 0! Caller should've checked...");
-  AspectRatio ratio = aAxisTracker.IsMainAxisHorizontal()
-                          ? aIntrinsicRatio
-                          : aIntrinsicRatio.Inverted();
+  AspectRatio ratio = aAxisTracker.IsMainAxisHorizontal() ?
+                      aIntrinsicRatio :
+                      aIntrinsicRatio.Inverted();
 
   return ratio.ApplyTo(aCrossSize);
 }
--- a/layout/generic/nsFlexContainerFrame.cpp
+++ b/layout/generic/nsFlexContainerFrame.cpp
@@ -557,8 +557,8 @@ public:
     return mFlexShrink * mFlexBaseSize;
   }
 
-  const nsSize& IntrinsicRatio() const { return mIntrinsicRatio; }
-  bool HasIntrinsicRatio() const { return mIntrinsicRatio != nsSize(); }
+  const AspectRatio IntrinsicRatio() const { return mIntrinsicRatio; }
+  bool HasIntrinsicRatio() const { return !!mIntrinsicRatio; }
 
   // Getters for margin:
   // ===================
@@ -756,7 +756,7 @@ protected:
   const float mFlexGrow;
   const float mFlexShrink;
 
-  const nsSize mIntrinsicRatio;
+  const AspectRatio mIntrinsicRatio;
 
   const nsMargin mBorderPadding;
   nsMargin mMargin; // non-const because we need to resolve auto margins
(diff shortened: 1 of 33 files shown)


mattatobin:
You need to fix the commit message for the second in this PR

RealityRipple:
I did not mean for the third one to get pushed yet, either... I'll clean it up as soon as the compile I'm working on is done.

wolfbeast:
The dimension computation was likely added by Mozilla in a different bug. It can probably be found by checking Mozilla's `gecko-dev` blame, but since this is a logical thing to do (set the missing value if the 2 dependent values exist) I don't see the point in spending the extra time.

layout/generic/nsFlexContainerFrame.cpp
```diff
@@ -1520,22 +1520,20 @@ CrossSizeToUseWithRatio(const FlexItem& aFlexItem,
...
-                        const nsSize& aIntrinsicRatio,
+                        const AspectRatio& aIntrinsicRatio,
                         const FlexboxAxisTracker& aAxisTracker)
 {
-  MOZ_ASSERT(aAxisTracker.GetCrossComponent(aIntrinsicRatio) != 0,
+  MOZ_ASSERT(aIntrinsicRatio,
              "Invalid ratio; will divide by 0! Caller should've checked...");
+  AspectRatio ratio = aAxisTracker.IsMainAxisHorizontal()
+                          ? aIntrinsicRatio
+                          : aIntrinsicRatio.Inverted();
```
wolfbeast, line 1534: Nit: code style
```
somelongvariable = somelongexpression ?
                   somevalue1 :
                   somevalue2;
```

layout/generic/nsHTMLCanvasFrame.cpp
```diff
@@ -211,10 +196,15 @@ nsHTMLCanvasFrame::GetIntrinsicSize()
...
-/* virtual */ nsSize
+/* virtual */ AspectRatio
 nsHTMLCanvasFrame::GetIntrinsicRatio()
 {
-  return IntrinsicRatioFromCanvasSize(GetCanvasSize());
+/*
+  if (StyleDisplay()->IsContainSize()) {
+    return AspectRatio();
+  }
+ */
```
wolfbeast, line 206: Please add a comment that this needs to be fixed when containsize is implemented.

layout/generic/nsVideoFrame.cpp
```diff
@@ -533,7 +531,9 @@ nsVideoFrame::ComputeSize(nsRenderingContext *aRenderingContext,
   intrinsicSize.height.SetCoordValue(size.height);
 
   // Only video elements have an intrinsic ratio.
-  nsSize intrinsicRatio = HasVideoElement() ? size : nsSize(0, 0);
+  auto intrinsicRatio = HasVideoElement()
+                            ? AspectRatio::FromSize(size)
+                            : AspectRatio();
```
wolfbeast, line 536: Nit: code style

layout/xul/nsImageBoxFrame.cpp
```diff
@@ -402,12 +402,13 @@ nsImageBoxFrame::PaintImage(nsRenderingContext& aRenderingContext,
...
     IntrinsicSize intrinsicSize;
-    nsSize intrinsicRatio;
+    AspectRatio intrinsicRatio;
     if (mIntrinsicSize.width > 0 && mIntrinsicSize.height > 0) {
       // Image has a valid size; use it as intrinsic size & ratio.
       intrinsicSize.width.SetCoordValue(mIntrinsicSize.width);
       intrinsicSize.height.SetCoordValue(mIntrinsicSize.height);
-      intrinsicRatio = mIntrinsicSize;
+      intrinsicRatio =
+          AspectRatio::FromSize(mIntrinsicSize);
```
wolfbeast, line 411: Nit: please make one line

layout/generic/nsHTMLCanvasFrame.cpp
RealityRipple: I figured 'contain: size' would be enough to search for in the future, but if there's a better label or "TODO:" prefix you'd like me to add, let me know. There's gonna be a lot more of these in the next PR.

wolfbeast: That comment is fine to copypasta on all commented-out blocks as far as I am concerned.

Issue #1610, opened 2020-07-19
Mozillaisms and other redundant settings that can be removed.


Peacock365:
Dear developers,

in the following a list of about:config settings that should be removed (or corrected) in my opinion, as of Pale Moon 28.11.0. Please forgive me my ignorance if I made any mistake here.

**Telemetry / crash reporter / Firefox experiments**

_toolkit.telemetry.unified
toolkit.telemetry.server_owner
toolkit.telemetry.server
toolkit.telemetry.reportingpolicy.firstRun
toolkit.telemetry.infoURL
toolkit.telemetry.enabled
toolkit.telemetry.debugSlowSql
devtools.telemetry.tools.opened.version
devtools.telemetry.supported_performance_marks
dom.ipc.plugins.flash.subprocess.crashreporter.enabled
dom.ipc.plugins.reportCrashURL
network.allow-experiments_

**Mozillaisms**

_browser.addon-watch.ignore
devtools.devedition.promo.url
devtools.devices.url
dom.push.serverURL
media.gmp-manger.url
media.gmp-manager.certs.1.commonName
media.gmp-manager.certs.2.commonName
services.sync.fxa.privacyURL
services.sync.fxa.termsURL
services.sync.tokenServerURl_

Again, forgive me my ignorance if I listed legitimate entries here that need to stay that way, this is just a post aimed at listing settings that might be redundant / overseen. I also don't think that Pale Moon / Basilisk still has active telemetry, crash reporter or something, I am sure most of these settings don't do anything anymore, but if that is the case, you might consider removing them to avoid confusion. Thank you for your attention.

mattatobin:
Just stop making irrelevant noise.

Peacock365:
@mattatobin 

How about you clean up redundant about:config entries to avoid confusion, instead of just closing this comment with a snarky remark?

@wolfbeast, I leave this to you, I know Tobin from the forums and I refuse to communicate with him.

mattatobin:
This issue tracker is not for you. It is for us. You can ask your questions at the forum. You have no choice who interacts with you here, like it or not UXP is as much mine as it is Moonchild's or Travis's or athenian200's or Ryan's or anyone else's who helped create and who helps maintain it.

I am telling you that this issue noise is unacceptable, the issue is not solicited, and this is most decently not one issue at all. It is innane rambling and I ask you to stop and head back to the forum to discuss these bits and bobs and WE will decide what and if any single bit gets made into an issue to be resolved.

[...]

PR #1634, opened 2020-08-15
No longer expose crash reporter settings to users.

Peacock365:
Partially resolves https://github.com/MoonchildProductions/UXP/issues/1610

### DIFF
--- a/modules/libpref/init/all.js
+++ b/modules/libpref/init/all.js
@@ -2839,9 +2839,6 @@ pref("dom.ipc.tabs.shutdownTimeoutSecs", 0);
 
 pref("dom.ipc.plugins.flash.disable-protected-mode", false);
 
-pref("dom.ipc.plugins.flash.subprocess.crashreporter.enabled", true);
-pref("dom.ipc.plugins.reportCrashURL", true);
-
 // How long we wait before unloading an idle plugin process.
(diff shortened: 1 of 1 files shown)


mattatobin:
Stop.

Peacock365:
What is going on with you? Crash reporter no longer exists, those settings are COMPLETELY, I emphasize COMPLETELY redundant.

wolfbeast:
> Crash reporter no longer exists, those settings are COMPLETELY, I emphasize COMPLETELY redundant.

You didn't bother to do proper research. This PR is therefore incomplete and shouldn't just be tossed into the tree.
Just take one of these prefs:

http://xref.palemoon.org/moonchild-central/search?string=reportCrashURL&find=

leading to http://xref.palemoon.org/moonchild-central/source/basilisk/modules/PluginContent.jsm#1039

What do you think would happen with that code if the pref is removed and the caller (not having the default-supplied form of getting prefs) is not? It would break. You can't just pull this out without an issue signalling all UXP consumers because it will potentially affect all applications building on UXP, and may break them.

So terse responses from Tobin aside, it needs a lot more care to remove things if you want to contribute these kinds of changes.

mattatobin:
I'd prefer not using those keywords at all.

PR #1635, opened 2020-08-16
Intrinsic Aspect Ratio: Debug Follow up.

g4jc:
Newly introduced aspect-ratio property did not have CSS_PROP_LIST_EXCLUDE_INTERNAL defines. 

Causing the following assertion:

`!nsCSSProps::PropHasFlags(p, (1<<28)) (properties defined outside of #ifndef CSS_PROP_LIST_EXCLUDE_INTERNAL sections must not have the CSS_PROPERTY_INTERNAL flag), at ...layout/style/nsCSSProps.cpp:289`

This patch resolves the assertion by adding #ifndef around the aspect-ratio property.

### DIFF
--- a/layout/style/nsCSSPropList.h
+++ b/layout/style/nsCSSPropList.h
@@ -470,6 +470,7 @@ CSS_PROP_DISPLAY(
     kAppearanceKTable,
     CSS_PROP_NO_OFFSET,
     eStyleAnimType_Discrete)
+#ifndef CSS_PROP_LIST_EXCLUDE_INTERNAL
 CSS_PROP_POSITION(
     aspect-ratio,
     aspect_ratio,
@@ -481,6 +482,7 @@ CSS_PROP_POSITION(
     nullptr,
     offsetof(nsStylePosition, mAspectRatio),
     eStyleAnimType_None)
+#endif // CSS_PROP_LIST_EXCLUDE_INTERNAL
 CSS_PROP_DISPLAY(
     backface-visibility,
     backface_visibility,


mattatobin:
And this does what now? How does this help debugging?

g4jc:
It's not possible to debug right now due to the assertion forcing a SIGSEV.  See: [Backtrace](https://paste.centos.org/view/125b41ee)

This #ifndef essentially informs the debugger to ignore the fact [CSS_PROPERTY_INTERNAL](https://github.com/MoonchildProductions/UXP/blob/22c6728132677aee60062790abc208b2c14928f2/layout/style/nsCSSPropList.h#L478) is being used here. The other options is to not call an internal function in the first place, as it is typically reserved for `-moz` selectors, however @RealityRipple may be able to shed light on why that function is needed here or which patch it was borrowed from.

RealityRipple:
The code in question was entirely just a guess on my part. I asked specifically about it multiple times. Sorry for putting it in the wrong place/not wrapping it. It does not directly exist in the code I pulled the patch from because FF switched to a completely different CSS methodology by the time the changes were made.

The property exists to handle the intrinsic aspect ratio of images before they're loaded. The actual CSS property `aspect-ratio` is a future pull request that I'm not up for touching yet, but part of the process this time around involved storing it internally for use in frame drawing calculations. It's set in `dom/html/nsGenericHTMLElement.cpp` (lines 1483 to 1502) and has control over `mAspectRatio` elsewhere in the code via `layout/style/nsRuleNode.cpp` unless I misread the way things work... Again, this was all a guess. If it all needs checks for this EXCLUDE_INTERNAL flag, that might be a little messy, but there is a config preference to turn off the setting of that value that could be defaulted to false (or the whole If statement in `nsGenericHTMLElement` can be wrapped) to keep it from ever being set, which should leave the ratio as 0 which means no calculations will be done without any errors. The property is `layout.css.intrinsic-aspect-ratio.enabled`.

g4jc:
Thanks for clarifying. I've never messed with CSS engine prior either so I'm not overly familiar with how things are done here, however, I was able to conclude that this string only appears inside `nsCSSProps` and the assertion is only triggered in debug mode. As such I don't think the other #ifndefs will be needed, I've confirmed it is working in debug mode with the current PR.

See for reference other instances of this string:
http://xref.palemoon.org/moonchild-central/search?string=CSS_PROP_LIST_EXCLUDE_INTERNAL&find=

RealityRipple:
I have no idea how `ValueForAspectRatio` ends up actually being generated, but as long as nothing crashes when you visit the testcase page ([here](https://uploads.realityripple.com/Projects/artest/)), then I guess we're good. (Failing a couple of the tests would be expected, of course.)

mattatobin:
Well the bulk of my question is.. Is the assert actually correct. I mean if we want to PROPERLY debug and this chunk of code etc shouldn't debugging assertions reflect the actual code changes? I mean what is up with that etc. Just walling it off so an assert passes just seems like a REALLY SHITTY solution if real debugging is to be happening.. Right?

Else you are fixing a beeping smoke alarm by taking out the battery as it were.

RealityRipple:
Pretty sure it's all about whether the property shows up or not in lists of public CSS properties, right? The comments in the code say:

> Callers may define CSS_PROP_LIST_EXCLUDE_INTERNAL if they want to exclude internal properties that are not represented in the DOM (only the DOM style code defines this).

The aspect-ratio property should not be getting manipulated directly at this stage of implementation; the debugging should be done via manipulation of the width and height html values and the checking of the image frame size. Later on, if and when `aspect-ratio` becomes an actual supported CSS property, then it will need specific manipulation checks and all that fun stuff, and will no longer be defined as internal anyway.

As far as I can tell, the error exists specifically and only to make sure that things defined as internal function as internal.


I may, however, be entirely wrong about all of this. My knowledge of all of this is literally nearly exactly one month old.

wolfbeast:
Thanks for the follow-up. It is as luke said in the commit comment; the check fails in debug because an internal property is declared outside of a considitonal block for internal properties. the assert double-checks if the code was properly ifdefed, nothing more. A self-check.

mattatobin:
Ah ok.

PR #1637, opened 2020-08-26
Don't preload nomodule scripts when modules are enabled 

g4jc:
Lands nomodules work. Ref: BZ 1382020 + HTML Parser Regeneration.


mattatobin:
Was this even tested?

PR #1649, opened 2020-09-12
Fix usage of a macro in a cocoa widget

adeshkp:
The unintended usage doesn't seem to affect anything, but results in a compiler warning.

### DIFF
--- a/widget/cocoa/nsDragService.mm
+++ b/widget/cocoa/nsDragService.mm
@@ -222,7 +222,7 @@ nsDragService::ConstructDragImage(nsIDOMNode* aDOMNode,
 bool
 nsDragService::IsValidType(NSString* availableType, bool allowFileURL)
 {
-  NS_OBJC_BEGIN_TRY_ABORT_BLOCK;
+  NS_OBJC_BEGIN_TRY_ABORT_BLOCK_RETURN;
 
   // Prevent exposing fileURL for non-fileURL type.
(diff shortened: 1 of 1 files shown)


wolfbeast:
Looking at the function, this needs to be `NS_OBJC_END_TRY_ABORT_BLOCK_RETURN(false)` instead.

adeshkp:
I've always seen `NS_OBJC_END_TRY_ABORT_BLOCK_RETURN` used in conjunction with `NS_OBJC_BEGIN_TRY_ABORT_BLOCK_RETURN`. See: https://github.com/MoonchildProductions/UXP/blob/a219152bddead6b34f40ae86217f39fbda91aab6/widget/cocoa/nsPrintSettingsX.mm#L50

And `NS_OBJC_END_TRY_ABORT_BLOCK` always seems to be paired with `NS_OBJC_BEGIN_TRY_ABORT_BLOCK`. See:
https://github.com/MoonchildProductions/UXP/blob/a219152bddead6b34f40ae86217f39fbda91aab6/widget/cocoa/nsPrintSettingsX.mm#L25

wolfbeast:
Well, you'd want to change the start macro too then :) This seems to have been a mistake on Mozilla's side.
The thing is, you'd want to return a value from that function, not an implicit boolean (that is false which is why it worked before)

mattatobin:
In the future PLEASE use the UXP Commit Message Style.

PR #1652, opened 2020-09-13
Clear the module map - Debug Followup

g4jc:
An assertion was added to dom/script/ScriptLoader.cpp but fails to compile on debug builds due it not being included in the namespace. This adds the missing namespace so that debug builds work again.


wolfbeast:
Could have sworn I added that one. Ohwell.

Issue #1647, opened 2020-09-11
Implement percentage for CSS opacity keywords


wolfbeast:
We currently support the standard float values for various `opacity` CSS keywords, but the spec has been changed to also permit percentages.
We should add parsing of percentages for `opacity` keywords to translate/serialize to the corresponding float value when read (as a percentage of "1")

This would involve changing `nsCSSPropList.h` to use `VARIANT_HPN` instead of `VARIANT_HN`, and adding or changing serialization from percentage to float for it. I haven't exactly found where this serialization happens for Opacity, yet, so that will require some searching.

athenian200:
Well, I've managed to get the style system to support opacity as a percentage without a problem. I just added support for that case into nsRuleNode.cpp and everything worked flawlessly from a visual perspective. Apparently we actually had support for this in some form a long time ago and it was simply removed.

The problem is that... well, as far as I can tell, if you change nsCSSPropList.h to VARIANT_HPN, the value _will_ serialize as a percentage, and if you don't change it, then percentages will be rejected as valid values right out of the gate. I could be wrong here, but the style system doesn't appear to support the concept of having a data type serialize as something different than what the user inputted, because there's a very rigid syntax in place designed to automate the adding of new elements and this violates the assumptions that framework makes. This can be changed of course, the hard part would be changing this for opacity without also changing it for the whole style system.

While you can easily convert between data type representations immediately after serialization, tests will fail because it technically serialized as a percentage even though the style system is actually converting the percentage to a float before anything happens internally.

wolfbeast:
Hmm... but the same thing happens with using percentages as alpha in `rgba()` -- that _does_ serialize to the float equivalent while the user has put in a percent value. So where is that conversion done then?

Or maybe the tests just need to be changed, or what do you mean, exactly?

athenian200:
> Hmm... but the same thing happens with using percentages as alpha in `rgba()` -- that _does_ serialize to the float equivalent while the user has put in a percent value. So where is that conversion done then?
> 

[...]

PR #1663, opened 2020-09-30
Remove excessive VARIANT_OPACITY statements.

athenian200:
I included VARIANT_OPACITY in too many places because I wanted it to be treated exactly like VARIANT_NUMBER for most purposes. This had some unintended consequences, but I've addressed it now.

Tag #1647


wolfbeast:
Verified. Merging.

